PR #6, opened 2015-01-31
Extended tutorial to show how to use `inshell`/`inproc`.  Fixes #4

### LINKED ISSUE #4: Recommended Example: Piping data between coreutils/other programs
Like restoring a database from a compressed database dump:

```bash
zcat file | mysql -uroot mydb
```

This is streamed right? A problem I had with shelly was -|- tried to read everything in at once and the api hard enough for me to forget about it and get busy on something else.

As an aside my use case is to do a few conditional checks/get user input with Haskell and then do mostly dumb piping between applications to update databases from different servers and other tasks like it.

Gabriel439:


Gabriel439:
@codygman: Check this out to see if this is the sort of example you had in mind

codygman:
Will do when I get back to a PC!

PR #11, opened 2015-02-02
Add example of restoring database from remote backup

codygman:
Does this look alright?


Gabriel439:
Try compiling the program with `-threaded` and see if that fixes it

PR #14, opened 2015-02-03
Fix mismatched parentheses in stdout function

MaxGabriel:


Gabriel439:
Thanks for fixing that!

PR #17, opened 2015-02-04
Fix #16: create a Format specifier for filepaths.

### LINKED ISSUE #16: Add `toTextIgnore`
I think it would be convenient to add the [shelly `toTextIgnore`](http://hackage.haskell.org/package/shelly-1.6.0/docs/src/Shelly-Base.html#toTextIgnore) function.

PierreR:


PierreR:
Done. That was indeed such a stupid position ;-)

Gabriel439:
Looks good.  Thanks!

PR #20, opened 2015-02-09
Fix typo in Shell `foldIO` law

happy4crazy:
Presumably we ought to be initializing the accumulator with `begin` rather than `step` :)


Gabriel439:
Oops!  Thank you for fixing that.  :)

PR #30, opened 2015-02-26
Fix haddock for fp

PierreR:
I have messed the haddock for `repl`. Sorry about that. This is the fix.

Do you think it worths adding a doctest ? 

```
>>> format fp ("~/.test.txt" ::FilePath)
"~/.test.txt"
```
or just:
```
>>> format fp "~/.test.txt"
"~/.test.txt"
```

### DIFF
--- a/src/Turtle/Format.hs
+++ b/src/Turtle/Format.hs
@@ -180,7 +180,7 @@ g = makeFormat (\n -> pack (showGFloat (Just 6) n ""))
 s :: Format r (Text -> r)
 s = makeFormat id
 
-{-| `Format` a `Filesystem.Path.CurrentOS` into `Text`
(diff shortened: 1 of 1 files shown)


Gabriel439:
No problem at all! :)

Yeah, I think it's worth having a doctest.  Maybe something like:

    >>> format fp ("usr" </> "lib")
    "usr/lib"

PierreR:
To my surprise I need to add the import `import Filesystem.Path.CurrentOS((</>))` (I can't import Turtle right there). Do you see why ?

Gabriel439:
Yeah, that's fine.  I'm okay with importing something in the doctest.

You probably can't import `Turtle` because then it would be a dependency cycle.

PR #36, opened 2015-03-20
Relax the return type of exit

maoe:
`exit` doesn't need to return a unit. This change is useful when you write something like below:

```haskell
action :: IO a
action = doSomething
  `catch` \e -> do
    doSomethingElse e
    exit 1 -- this doesn't type check!

doSomething :: IO a
```


Gabriel439:
Good catch!

Thank you for fixing that.  I agree that it's way more useful with the polymorphic return value.

maoe:
Thanks!

Issue #32, opened 2015-03-10
Capturing shell output and exit status?


houshuang:
Maybe I'm misunderstanding, but it seems to me that when running an external shell command, I can choose between capturing either the exit code, through `shell`, which prints the output and returns the exit code, or capturing the output through `inproc`, which returns the output as a stream, but not the exit code? Instead of `ExitCode`, could we do something like `data Result = ExitSuccess Shell Text | ExitFailure Int`? So that I can pattern match and say if success, do something with the output, if failure, do something else... It could also return a stream with the exit code I guess.

It seems like you might be able to wrap `shell` with `liftIO` to get this effect?

Gabriel439:
There wouldn't be much advantage to returning a `Shell Text` (as opposed to just `Text`) because you would have to buffer all of standard output until you knew whether the command exited successfully or not.

houshuang:
OK, so Shell Text is a stream, that makes sense (sorry, just learning about this library). 

In this particular case, I wanted to do `git ls-files` on a directory and capture the output if it succeeded, and run `find . -type d` if it failed.

Gabriel439:
In that case what you probably want is to store the entire output in the result:

```haskell
data Result = ExitSuccess Text | ExitFailure Int
```

... where that `Text` is the entire standard output contents (i.e. every single line).

There's no good streaming way to do what you want, because otherwise your code would behave incorrectly in the corner case where `git ls-files` produces some output and then fails in the middle (which is unlikely, but possible).

At this point, the hardest part is just coming up with a name for this variation on `shell`/`proc`.

As a side note, `System.Process` has a function that already does this named `readProcessWithExitCode` ([Link](http://hackage.haskell.org/package/process-1.2.0.0/docs/System-Process.html#v:readProcessWithExitCode)), except it uses `String` instead of `Text`.

Gabriel439:
This was fixed by #42

PR #42, opened 2015-05-01
Add functions that return Exitcodes and Stdout

Profpatsch:
Possible fix for #32.
Here goes nothing.

The question is: Should there be something about `stderr`, too? If so, maybe there is a better way to simplify `createProcess` from without adding API-Overhead for every possible use case.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -98,6 +98,8 @@ module Turtle.Prelude (
     -- * IO
       proc
     , shell
+    , procCollect
+    , shellCollect
     , echo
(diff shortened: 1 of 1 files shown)


Gabriel439:
This will work correctly in the face of exceptions, meaning that if either computation throws an exception the other one will be canceled.

The only thing I might suggest is to name these utilities`procStrict` and `shellStrict` for consistency with the `strict` function.  Other than that they look good.

Profpatsch:
Done, compiles.

Gabriel439:
Thanks!

Issue #43, opened 2015-05-03
Does it make sense to instance Num Shell?


Profpatsch:
I’m fixing build warnings atm and noticed that there is

``` haskell
instance Monoid a => Num (Shell a) where
    fromInteger n = select (replicate (fromInteger n) mempty)

    (+) = (<|>)
    (*) = (<>)
```

It’s missing `abs`, `signum` and `-`. Do these even make sense?
And if not, why instance it at all. To make it more beginner-friendly? But then again that reminds me of C++ Operator Overloading.

Profpatsch:
It still compiles w/o the instance at least.

Profpatsch:
Same with `Pattern`.

Gabriel439:
It's because `Shell`s form a semiring, but Haskell has no semiring class an `Num` is the closest approximation

Same thing for `Pattern`

Profpatsch:
Hah, I see. More abstract algebra I should remember from university. :P

Profpatsch:
How about using https://hackage.haskell.org/package/semiring-simple-0.2.1.0/docs/Data-Semiring.html ?

Gabriel439:
The biggest issue with that package is the use of overlapping instances like:

```haskell
    instance Semiring a => Monoid a where
```

That will conflict with every `Monoid` instance, even for types that do not implement `Semiring` (unless Haskell has changed its typeclass resolution behavior since the last time I checked).

The second reason is that you lose support for overloaded numeric literals if you use the `Semiring` class.  I like being able to do things like:

```
>>> view (3 * "Example)
"Example"
"Example"
"Example"
```

Profpatsch:
Ah, so it’s best to use the builtins.

Gabriel439:
Is this still an issue for you?

Profpatsch:
Nope.

PR #44, opened 2015-05-03
Add documentation to Num instances.

Profpatsch:
#43


Profpatsch:
This should fix the warnings and document the Semiring case.

Profpatsch:
Okay, it seems the new GHC changed some import strategies, Travis says this doesn’t build.

Gabriel439:
It's because those imports are necessary for older versions of `ghc` (pre-7.10)

Gabriel439:
Thanks!

PR #46, opened 2015-05-05
Generalize IO to MonadIO for many functions

### LINKED ISSUE #45: Nicer Shell and IO integration
It may be frustrating that functions such as `testfile` can't be called in the `Shell` monad. Yes, I can apply `liftIO`, but wouldn't it make sense to cut the noise and do it in the library, so `testfile` would have the following signature?
```haskell
testfile :: MonadIO m => FilePath -> m Bool
```

It could be run seamlessly from both `IO` and `Shell`. Same for other functions.

int-index:
closes #45

Please review carefully.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -171,7 +171,7 @@ import Control.Concurrent.Async (Async, withAsync, wait, concurrently)
 import Control.Concurrent (threadDelay)
 import Control.Exception (bracket, throwIO)
 import Control.Foldl (FoldM(..), list)
-import Control.Monad (msum)
+import Control.Monad (liftM, msum)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Managed (Managed, managed)
 #ifdef mingw32_HOST_OS
@@ -720,8 +720,8 @@ append file s = sh (do
     liftIO (Text.hPutStrLn handle txt) )
 
 -- | Read in a stream's contents strictly
-strict :: (MonadIO io, Functor io) => Shell Text -> io Text
-strict s = fmap Text.unlines (fold s list)
+strict :: MonadIO io => Shell Text -> io Text
+strict s = liftM Text.unlines (fold s list)
 
 -- | Acquire a `Managed` read-only `Handle` from a `FilePath`
 readonly :: FilePath -> Managed Handle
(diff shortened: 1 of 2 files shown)


src/Turtle/Prelude.hs
```diff
@@ -233,11 +235,12 @@ proc cmd args = system (Process.proc (unpack cmd) (map unpack args))
...
 -}
 shell
-    :: Text
+    :: MonadIO io
+    => Text
     -- ^ Command line
     -> Shell Text
     -- ^ Lines of standard input
-    -> IO ExitCode
+    -> io ExitCode
```
Gabriel439: Oh that's clever!  I like that.  :)

src/Turtle/Shell.hs
```diff
@@ -78,18 +79,22 @@ import Data.Monoid (Monoid(..), (<>))
 import Data.String (IsString(..))
 
 -- | A @(Shell a)@ is a protected stream of @a@'s with side effects
-newtype Shell a = Shell { foldIO :: forall r . FoldM IO a r -> IO r }
+newtype Shell a = Shell { _foldIO :: forall r . FoldM IO a r -> IO r }
+
+-- | Use a `FoldM IO` to reduce the stream of @a@'s produced by a `Shell`
```
Gabriel439: Put `foldIO` below `fold` to suggest to people `fold` is the more commonly used function

Gabriel439: Actually, I take that back.  It would be inconsistent with the rest of the progression (each new function is implemented in terms of the previous one).

src/Turtle/Prelude.hs
```diff
@@ -367,61 +374,61 @@ stream p s = do
     inhandle hOut
 
 -- | Print to @stdout@
-echo :: Text -> IO ()
-echo = Text.putStrLn
+echo :: MonadIO io => Text -> io ()
+echo t = liftIO (Text.putStrLn t)
```
Gabriel439: For this and `err`, name the parameter `txt`

Gabriel439:
Looks good!  Just a few cosmetic suggestions, but otherwise I'm ready to merge this.

src/Turtle/Prelude.hs
int-index: I looked up parameter names in corresponding libraries. `text` calls the parameter `t` in the definition of `Text.putStrLn`. However, I too think that it's not as descriptive as it should be. I'll change that in a minute.

int-index:
Done. One thing that caught my attention is that the `Functor io` constraint on the `strict` function is redundant when compiling with GHC 7.10. I'm not sure if it needs to be conditionally removed with CPP.

Gabriel439:
One last thing: use liftM or do notation to remove the Functor constraint for pre-7.10 ghc

Gabriel439:
Looks good!  Give me some time to update the documentation before I upload this to Hackage.

PR #48, opened 2015-05-07
Add `chmod` and related definitions

### LINKED ISSUE #47: chmod: change file mode bits
I needed to make a file executable, so I've implemented `chmod` as follows:

```haskell
chmod :: MonadIO io => FilePath -> (Permissions -> Permissions) -> io ()
chmod path f = liftIO $ do
    let path' = encodeString path
    getPermissions path' >>= setPermissions path' . f
```

It can be used like so:
```haskell
chmod filepath (setOwnerExecutable True)
```

I'm not sure it's the best design because I have only one use-case. Any ideas to improve the definition to include it in the library?

int-index:
fixes #47

Only a draft, criticism is welcome.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -136,13 +136,13 @@ module Turtle.Prelude (
     , (.||.)
 
     -- * Permissions
-    , Directory.Permissions
+    , Permissions
     , chmod
     , readable, nonreadable
     , writable, nonwritable
     , executable, nonexecutable
     , searchable, nonsearchable
-    , ooo, roo, owo, oox, oos, rwo, rox, ros, owx, rwx, rws
+    , ooo,roo,owo,oox,oos,rwo,rox,ros,owx,rwx,rws
 
     -- * Managed
     , readonly
@@ -204,6 +204,7 @@ import System.Environment (
     lookupEnv,
 #endif
     getEnvironment )
+import System.Directory (Permissions)
 import qualified System.Directory as Directory
 import System.Exit (ExitCode(..), exitWith)
 import System.IO (Handle)
@@ -592,34 +593,29 @@ touch file = do
 > chmod executable "foo.txt"  -- chmod u+x foo.txt
 > chmod nonwritable "foo.txt" -- chmod u-x foo.txt
 -}
-chmod
-    :: MonadIO io
-    => (Directory.Permissions -> Directory.Permissions)
-    -> FilePath
-    -> io ()
+chmod :: MonadIO io => (Permissions -> Permissions) -> FilePath -> io ()
 chmod modifyPermissions path = liftIO (do
     let path' = deslash (Filesystem.encodeString path)
     permissions <- Directory.getPermissions path'
     Directory.setPermissions path' (modifyPermissions permissions) )
 
-readable, nonreadable :: Directory.Permissions -> Directory.Permissions
+readable, nonreadable :: Permissions -> Permissions
 readable = Directory.setOwnerReadable True
 nonreadable = Directory.setOwnerReadable False
 
-writable, nonwritable :: Directory.Permissions -> Directory.Permissions
+writable, nonwritable :: Permissions -> Permissions
 writable = Directory.setOwnerWritable True
 nonwritable = Directory.setOwnerWritable False
 
-executable, nonexecutable :: Directory.Permissions -> Directory.Permissions
+executable, nonexecutable :: Permissions -> Permissions
 executable = Directory.setOwnerExecutable True
 nonexecutable = Directory.setOwnerExecutable False
 
-searchable, nonsearchable :: Directory.Permissions -> Directory.Permissions
+searchable, nonsearchable :: Permissions -> Permissions
 searchable = Directory.setOwnerSearchable True
 nonsearchable = Directory.setOwnerSearchable False
 
-ooo, roo, owo, oox, oos, rwo, rox, ros, owx, rwx, rws
-    :: Directory.Permissions -> Directory.Permissions
+ooo,roo,owo,oox,oos,rwo,rox,ros,owx,rwx,rws :: Permissions -> Permissions
 ooo = const Directory.emptyPermissions
 roo = readable . ooo
 owo = writable . ooo
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -136,13 +136,13 @@ module Turtle.Prelude (
     , (.||.)
 
     -- * Permissions
+    , Directory.Permissions
     , chmod
-    , Permissions
     , readable, nonreadable
     , writable, nonwritable
     , executable, nonexecutable
     , searchable, nonsearchable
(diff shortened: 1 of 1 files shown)


src/Turtle/Prelude.hs
```diff
@@ -128,6 +128,13 @@ module Turtle.Prelude (
     , date
     , datefile
     , touch
+    , chmod
```
Gabriel439: It's probably worth giving `chmod` its own section since it introduces so many new names

src/Turtle/Prelude.hs
```diff
@@ -575,6 +588,44 @@ touch file = do
 #endif
         else output file empty )
 
+type Permissions = Filesystem.Permissions -> Filesystem.Permissions
```
Gabriel439: I would just skip the type synonym.  The fully expanded type is not that bad and it will avoid confusion

src/Turtle/Prelude.hs
```diff
@@ -195,7 +202,11 @@ import System.Environment (
     lookupEnv,
 #endif
     getEnvironment )
-import System.Directory (getPermissions, readable)
+import qualified System.Directory as Filesystem
+    ( Permissions, getPermissions, setPermissions, emptyPermissions
```
Gabriel439: What's the purpose of these explicit imports if you qualify everything anyway?

int-index: Conflict resolution. Some names from `System.Directory` clash with names from `Filesystem.Path.CurrentOS`, and I qualified `System.Directory` as `Filesystem`. Should I qualify it with another name?

src/Turtle/Prelude.hs
int-index: The fully expanded type is just too long for my taste (especially with qualified names) and it caused me to split type signatures into multiple lines to fit 80 characters, so I introduced a synonym. I can remove it, though, if you think it's confusing.

src/Turtle/Prelude.hs
```diff
@@ -575,6 +588,44 @@ touch file = do
 #endif
         else output file empty )
 
+type Permissions = Filesystem.Permissions -> Filesystem.Permissions
+
+-- | Update a file or directory's permissions
```
Gabriel439: Add a couple of example usages in the documentation side by side with the equivalent Bash:

```haskell
{-| Update a file or directory's permissions

> chmod rwo        "foo.txt"  -- chmod u=rw foo.txt
> chmod executable "foo.txt"  -- chmod u+x foo.txt
> chmod nonwritable "foo.txt" -- chmod u-x foo.txt
-}
```

src/Turtle/Prelude.hs
int-index: Why not? Done.

src/Turtle/Prelude.hs
Gabriel439: Yeah, just qualify it as `Directory` instead of `Filesystem`

src/Turtle/Prelude.hs
Gabriel439: I generally try to avoid type synonyms, mainly because they are not preserved well by type inference and by type errors.  Every time I introduce a type synonym it always bites me in the end.  They are an ongoing support burden (very common source of questions on mailing lists and stack overflow) and poor new user experience (due to poor type inference and type errors).

If you really would like to compress the type then I would recommend wrapping it in a newtype.

int-index: I'll just remove it then.

int-index:
Here's an idea: if we return the updated permissions, they could be examined by calling `chmod id`. What do you think?

Gabriel439:
For consistency with Unix, I'd probably leave that for a separate `stat` command

int-index:
`chmod` has a `-v` option that allows me to see if the mode was changed, what it was and what it became. So it'd make sense to return something.

Gabriel439:
Oh, I didn't realize that `chmod` could do that!  In that case, that makes a lot of sense to return the permissions from `chmod`

int-index:
Any ideas to further improve the definition? Or can it be merged?

Gabriel439:
Looks good!  Thanks for writing that up!

Gabriel439:
One last thing before I release this.  I was thinking of changing `chmod` slightly to not return the `Bool` value indicating whether or not the value was changed.  My reasoning is that (A) chmod doesn't actually tell whether the file changed, (B) the user can easily do this test themselves just by using `(/=)`, and (C) the `Bool` is sort of weakly typed; you don't really know what it means unless you read the documentation.

int-index:
I don't have a strong opinion on the topic, but here are some thoughts.

(A) Agreed, but it's strictly an improvement.
(B) That would require more IO operations, thus hurt performance. I don't have any benchmarks, but it was my main motivation for adding this `Bool` value. Alternatively, we could return both old and new permissions (but it doesn't fit what the user could expect from `chmod`).
(C) Agreed, but that's probably better than a specialized type until we get a proper `Boolish` typeclass.

So if you don't care that in some rare cases users would be forced to perform more IO than necessary (hurting the performance), feel free to remove this `Bool` value.

Possible solution: define `chmod'` that returns the `Bool` value and `chmod` that doesn't.

int-index:
Given that `chmod -v` returns both old and new permissions, I think we should just change that to `(Permissions, Permissions)`.

Gabriel439:
Oh yeah, I forgot that the user wouldn't have the old permissions.  Is there any Unix utility equivalent to `Directory.getPermissions` that we could add?  It seems like it would be more straightforward for the user to get the old permissions directly instead of indirectly through `chmod`

int-index:
I suppose `stat` is used for that purpose.

Gabriel439:
Then how about `chmod` just returns a single `Permissions` for now and we save the retrieval of the old permissions for a future `stat` function?

int-index:
I'm fine with it, but there are still two counter-arguments:

1. `chmod -v` returns both old and new permissions, so should we.
2. The `chmod` function performs `Directory.getPermissions`, which is a disk IO operation. What if it's a network file system so each operation has a big cost? If perform IO, we better not throw the result away.

int-index:
So I'd rather return `(old, new) :: (Permissions, Permission)`, but I don't insist.

Gabriel439:
What if we were to deviate from the Unix model slightly and provide `getmod` and `setmod` in addition to `chmod`?

```haskell
getmod :: IO Permissions
setmod :: Permissions -> IO Permissions
chmod  :: (Permissions -> Permissions) -> IO Permissions
```

Then if the user wanted to update permissions while retrieving the old permissions using minimal `IO` operations, they would write:

```haskell
oldPerm <- getmod
newPerm <- setmod (rwo oldPerm)
```

3noch:
Having a getter and setter would also be useful when the modification function is not pure, as `chmod` requires.

int-index:
@Gabriel439 Yes, this solution seems to address all raised concerns.

Gabriel439:
Alright I added `getmod` and `setmod` and uploaded the most recent changes to Hackage (including the `MonadIO` generalizations and a corresponding section in the tutorial)

PR #51, opened 2015-05-11
Command line arguments

### LINKED ISSUE #50: How about a simple getArgs?
I wrote the following script as a Turtle learning exercise and was pleased with how it turned out (after a lot of backing and filling over discovering Turtle features and figuring out what else had to be imported).  Supporting
Optparse-Applicative is a wonderful goal, but a simple addition of getArgs :: IO [Text] would go a long way towards cleaning up simple scripts.

    #! /usr/bin/env runghc

    {-# LANGUAGE OverloadedStrings #-}

    import qualified Data.Text as T
    import System.Environment (getArgs)
    import Turtle

    -- Search for haskell files using a given symbol a la unix grep.

    main = do
      [symS] <- getArgs
      let sym = T.pack symS
      dir <- fmap (</> "hs") home
      sh (do file <- find (suffix ".hs") dir
             let fn = dirname file </> basename file
             m <- grep (has (text sym)) (input file)
             liftIO . echo $ format (fp % ": " % s) fn m
         )

int-index:
fixes #50

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -103,6 +103,7 @@ module Turtle.Prelude (
     , echo
     , err
     , readline
+    , arguments
 #if MIN_VERSION_base(4,7,0)
     , export
     , unset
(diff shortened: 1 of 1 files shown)


Gabriel439:
Thanks!

PR #56, opened 2015-05-14
Turtle.Options module

### LINKED ISSUE #13: Provide easy to use arguments parsing
It looks like Turtle does not provide arguments parsing facilities. I'd love to see something similar to optparse integrated in easy-to-use way into Turtle

int-index:
This is a place to discuss the new `Turtle.Options` module. It's meant to be a simplified interface to the `optparse-applicative` library that allows to quickly add options parsing to a script. It's not meant to give fine-grained control or much power over the parsing process, for those purposes the `optparse-applicative` library should be used directly.

This pull request is not meant to be merged until the module is completed.

fixes #13

### DIFF
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -40,6 +40,14 @@ instance Option Text where
     type OptionFields Text = Opts.ArgumentFields
     opt = Opts.argument (fmap Text.pack Opts.readerAsk)
 
+instance Option Int where
+    type OptionFields Int = Opts.ArgumentFields
+    opt = Opts.argument Opts.auto
+
+instance Option Integer where
+    type OptionFields Integer = Opts.ArgumentFields
+    opt = Opts.argument Opts.auto
+
 flag :: a -> a -> Opts.Mod Opts.FlagFields a -> Opts.Parser a
 flag = Opts.flag
 
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -19,6 +19,7 @@ module Turtle.Options
     ) where
 
 import Data.Monoid
+import Data.Foldable
 import Data.String (IsString)
 import Text.Read (readMaybe)
 import Data.Text (Text)
@@ -42,14 +43,8 @@ newtype ParameterName = ParameterName { getParameterName :: Text }
 newtype LongName = LongName { getLongName :: Text }
     deriving (IsString)
 
-newtype ShortName = ShortName { getShortName :: Char }
-
 newtype HelpMessage = HelpMessage { getHelpMessage :: Text }
     deriving (IsString)
-
-optionalOr :: Monoid m  => (a -> m) -> Optional a -> m
-optionalOr _ Default = mempty
-optionalOr f (Specific a) = f a
 
 switch
     :: LongName
@@ -58,8 +53,8 @@ switch
 switch longName helpMessage
    = Opts.switch
    $ (Opts.long . Text.unpack . getLongName) longName
-  <> maybe mempty (Opts.short . fst) (Text.uncons (getLongName longName))
-  <> optionalOr (Opts.help . Text.unpack . getHelpMessage) helpMessage
+  <> foldMap (Opts.short . fst) (Text.uncons (getLongName longName))
+  <> foldMap (Opts.help . Text.unpack . getHelpMessage) helpMessage
 
 parameter
     :: ParameterRead a
@@ -69,7 +64,7 @@ parameter
 parameter paramRead paramName helpMessage
    = Opts.argument (parameterReadToReadM paramRead)
    $ Opts.metavar (Text.unpack (getParameterName paramName))
-  <> optionalOr (Opts.help . Text.unpack . getHelpMessage) helpMessage
+  <> foldMap (Opts.help . Text.unpack . getHelpMessage) helpMessage
 
 newtype ParameterRead a = ParameterRead (ReaderT String Maybe a)
     deriving (Functor, Applicative, Monad)
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -15,6 +15,8 @@ module Turtle.Options
     , parameter
     , pAuto
     , pText
+    , pInteger
+    , pDouble
     ) where
 
 import Data.Monoid
@@ -22,6 +24,7 @@ import Data.String (IsString)
 import Text.Read (readMaybe)
 import Data.Text (Text)
 import qualified Data.Text as Text
+import Data.Optional
 import Control.Applicative
 import Control.Monad.IO.Class
 import Control.Monad.Trans.Reader
@@ -45,30 +48,30 @@ newtype ShortName = ShortName { getShortName :: Char }
 newtype HelpMessage = HelpMessage { getHelpMessage :: Text }
     deriving (IsString)
 
+optionalOr :: Monoid m  => (a -> m) -> Optional a -> m
+optionalOr _ Default = mempty
+optionalOr f (Specific a) = f a
+
 switch
     :: LongName
-    -> ShortName
-    -> HelpMessage
+    -> Optional ShortName
+    -> Optional HelpMessage
     -> Parser Bool
 switch longName shortName helpMessage
    = Opts.switch
-   $ Opts.long (Text.unpack (getLongName longName))
-  <> Opts.short (getShortName shortName)
-  <> Opts.help (Text.unpack (getHelpMessage helpMessage))
+   $ (Opts.long . Text.unpack . getLongName) longName
+  <> optionalOr (Opts.short . getShortName) shortName
+  <> optionalOr (Opts.help . Text.unpack . getHelpMessage) helpMessage
 
 parameter
     :: ParameterRead a
     -> ParameterName
-    -> LongName
-    -> ShortName
-    -> HelpMessage
+    -> Optional HelpMessage
     -> Parser a
-parameter paramRead paramName longName shortName helpMessage
+parameter paramRead paramName helpMessage
    = Opts.option (parameterReadToReadM paramRead)
    $ Opts.metavar (Text.unpack (getParameterName paramName))
-  <> Opts.long (Text.unpack (getLongName longName))
-  <> Opts.short (getShortName shortName)
-  <> Opts.help (Text.unpack (getHelpMessage helpMessage))
+  <> optionalOr (Opts.help . Text.unpack . getHelpMessage) helpMessage
 
 newtype ParameterRead a = ParameterRead (ReaderT String Maybe a)
     deriving (Functor, Applicative, Monad)
@@ -81,6 +84,15 @@ pAuto = ParameterRead (ReaderT readMaybe)
 
 pText :: ParameterRead Text
 pText = ParameterRead (ReaderT $ \s -> Just (Text.pack s))
+
+pInteger :: ParameterRead Integer
+pInteger = pAuto
+
+pInt :: ParameterRead Int
+pInt = pAuto
+
+pDouble :: ParameterRead Double
+pDouble = pAuto
 
 parameterReadToReadM :: ParameterRead a -> Opts.ReadM a
 parameterReadToReadM (ParameterRead f) = do
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -1,70 +1,90 @@
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 -- TODO: documentation
 
 module Turtle.Options
-    ( Opts.Parser
-    , Opts.InfoMod
+    ( Parser
+    , ParameterRead
+    , parameterRead
+    , ParameterName(..)
+    , LongName(..)
+    , ShortName(..)
+    , HelpMessage(..)
     , options
-    , opt
-    , flag
-    , metavar
-    , helpmsg
-    , shortname
-    , longname
-    , header
-    , footer
+    , switch
+    , parameter
+    , pAuto
+    , pText
     ) where
 
+import Data.Monoid
+import Data.String (IsString)
+import Text.Read (readMaybe)
 import Data.Text (Text)
 import qualified Data.Text as Text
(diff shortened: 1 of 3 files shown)


int-index:
Here's an example of how this module can be used
```haskell
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE RecordWildCards #-}

import Turtle

data Options = Options
    { name :: Text
    , age :: Integer
    , excited :: Bool }

pOptions = Options
    <$> opt (metavar "NAME" <> helpmsg "Your first name")
    <*> opt (metavar "AGE"  <> helpmsg "Your current age")
    <*> opt (shortname 'e' <> longname "excited" <> helpmsg "Are you excited?")

main :: IO ()
main = do
    Options{..} <- options pOptions
        (header "Greeting script" <> footer "Turtle rocks")
    let exc = if excited then "!!!" else ""
    echo ("Hello there, " <> name <> exc)
    if age < 35
        then echo "You are young, so many opportunities! Writing shell scripts\
                  \ with Turtle is one of them!"
        else echo "You've lived long enough to know that static typing is\
                  \ better than no typing. Learn Turtle for shell scripting."
```

Any suggestions?

Gabriel439:
Give me a few days to review this.  Right now I'm preparing for a talk that I have to give.

3noch:
Links to video and/or slides are welcome. :wink:

int-index:
Hi @Gabriel439 

It's been a little longer than a few days, so just in case you forgot about this PR, I want to remind you that it awaits review. I can't continue working on the feature until I have some feedback.

src/Turtle/Options.hs
```diff
@@ -0,0 +1,70 @@
...
+
+import Data.Text (Text)
+import qualified Data.Text as Text
+import Control.Applicative
+import Control.Monad.IO.Class
+import qualified Options.Applicative as Opts
+import qualified Options.Applicative.Types as Opts
+import qualified Options.Applicative.Builder.Internal as Opts
+
+options :: MonadIO io => Opts.Parser a -> Opts.InfoMod a -> io a
```
Gabriel439: Import the `Parser` type unqualified.  Some versions of haddock will show any qualifiers in the type signature

Gabriel439:
I'd recommend departing a little bit more from the `optparse-applicative` API, mainly because the types are a bit complex (Particularly the `HasMetaVar`/`HasName` type classes and the `Mod`/`InfoMod` types).

I was thinking of instead hard-coding the modifiers that a user must specify for each field, only because it would simplify the types a lot:

```haskell
newtype HelpMessage = HelpMessage Text deriving (IsString)
newtype LongName    = LongName Text deriving (IsString)
type    ShortName   = Char

bool ::            LongName -> ShortName -> HelpMessage -> Parser Bool
int  :: MetaVar -> LongName -> ShortName -> HelpMessage -> Parser Int
text :: MetaVar -> LongName -> ShortName -> HelpMessage -> Parser Text

newtype Header = Header Text deriving (IsString)
newtype Footer = Footer Text deriving (IsString)

options :: MonadIO a => Header -> Footer -> Parser a -> io a
```

You could even generalize `int` to:

```haskell
num :: Num n => MetaVar -> LongName -> ShortName -> HelpMessage -> Parser n
```

Also, maybe we might drop the `Footer` from `options` and just using `Header`:

```haskell
options :: MonadIO a => Header -> Parser a -> io a
```

int-index:
What if the user doesn't want to provide `LongName` or `ShortName` or `HelpMessage`? We could wrap all of them in `Maybe` or add `NoLongName` `NoShortName` `NoHelpMessage` constructors to the newtypes.

Gabriel439:
You could actually go one step further:

```haskell
data Optional a = Specific a | Unspecified

instance Alternative Optional where
    empty = Unspecified

    Unspecified <|> x = x
    x           <|> _ = x

instance Applicative Optional where
    pure = Specific

    Specific f <*> Specific x = Specific (f x)
    _          <*> _          = Unspecified

instance IsString a => IsString (Optional a)
    fromString str = Specific (fromString str)
```

Then the type would be something like:

```haskell
bool :: Optional LongName -> Optional ShortName -> Optional HelpMessage -> Parse Bool
```

Thanks to the `IsString` instance for `Optional` you could provide a string literal as an argument:

```haskell
bool "dynamic" (pure 'd') "enable dynamic linking"
```

... but if you wanted to leave them unspecified you can use `empty`:

```haskell
bool empty empty empty
```

Unfortunately, Haskell doesn't let you overload character literals, but I think that `pure` / `Specific` are okay to use for `Optional ShortName`.

int-index:
I suppose the only difference from `Maybe` is that we don't have to write an orphan `IsString` instance?

int-index:
@Gabriel439 

I've found a discussion in the Haskell libraries mailing list where you weren't very fond of the idea of adding such an instance to `base` (https://mail.haskell.org/pipermail/libraries/2013-July/thread.html#20366)

Did you change your opinion? If not, how is this case different?

Gabriel439:
The conclusion I eventually came to was that if you provide an `IsString` instance, then it should satisfy these laws:

```haskell
fromString (str1 <> str2) = fromString str1 <> fromString str2

fromString mempty = mempty
```

... which it would if we added this `Monoid` instance to `Optional`:

```haskell
instance Monoid a => Monoid (Optional a) where
    mempty = pure mempty
    mappend = liftA2 mappend
```

The problem with the `Maybe` type is that it has the wrong `Monoid` instance

int-index:
The `Option` type from the `semigroups` package has the right `Monoid` instance.

Gabriel439:
Also, it occurred to me that we may need to make either the `ShortName` or `LongName` required.  I'm actually not sure what `optparse-applicative` does if you omit both.

int-index:
If you specify neither `ShortName` nor `LongName`, `optparse-applicative` uses the default value (`False`) and when you run the program you can't change it.

I don't like this behavior. I think we should use the `These` type to specify `ShortName` and `LongName` (https://hackage.haskell.org/package/these-0.4.2/docs/Data-These.html)

Gabriel439:
Maybe it would be simpler if we just made the `LongName` required.  I think it's good practice to always provide a long name for a flag anyway

int-index:
I hesitate to introduce this `Optional` type. For the sake of simplification, I think it's acceptable to use `Maybe` without the `IsString` instance.

Gabriel439:
Alright.  What about making the `LongName` a required argument?

int-index:
Yes, I think it's a good idea — I can't imagine a situation in which a `LongName` would hurt, and the module isn't meant to give fine-graind control after all.

Gabriel439:
Alright, that sounds like a good plan.  Just use `Maybe`, then.

int-index:
I suppose now we can use your [optional-args](https://hackage.haskell.org/package/optional-args) library?

Gabriel439:
Yep!  That was the idea :wink:

int-index:
Updated example:
```haskell
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE RecordWildCards #-}

import Turtle
import Data.Optional

data Options = Options
    { name :: Text
    , age :: Integer
    , excited :: Bool }

pOptions = Options
    <$> parameter pText "NAME" "Your first name"
    <*> parameter pInteger "AGE" "Your current age"
    <*> switch "excited" "Are you excited?"

main :: IO ()
main = do
    Options{..} <- options "Greeting script" pOptions
    let exc = if excited then "!!!" else ""
    echo ("Hello there, " <> name <> exc)
    if age < 35
        then echo "You are young, so many opportunities! Writing shell scripts\
                  \ with Turtle is one of them!"
        else echo "You've lived long enough to know that static typing is\
                  \ better than no typing. Learn Turtle for shell scripting."
```

Short name for a parameter is derived automatically from the first letter of its long name.

src/Turtle/Options.hs
```diff
@@ -0,0 +1,100 @@
...
+    $ Opts.execParser
+    $ Opts.info (Opts.helper <*> parser) (Opts.header (Text.unpack header))
+
+newtype ParameterName = ParameterName { getParameterName :: Text }
+    deriving (IsString)
+
+newtype LongName = LongName { getLongName :: Text }
+    deriving (IsString)
+
+newtype ShortName = ShortName { getShortName :: Char }
```
Gabriel439: You can delete this since you don't use it any longer

src/Turtle/Options.hs
```diff
@@ -0,0 +1,100 @@
...
+
+newtype HelpMessage = HelpMessage { getHelpMessage :: Text }
+    deriving (IsString)
+
+optionalOr :: Monoid m  => (a -> m) -> Optional a -> m
```
Gabriel439: `optionalOr` is the exact same thing as `Data.Foldable.foldMap` since `Optional` implements `Foldable`

src/Turtle/Options.hs
```diff
@@ -0,0 +1,100 @@
...
+optionalOr f (Specific a) = f a
+
+switch
+    :: LongName
+    -> Optional HelpMessage
+    -> Parser Bool
+switch longName helpMessage
+   = Opts.switch
+   $ (Opts.long . Text.unpack . getLongName) longName
+  <> maybe mempty (Opts.short . fst) (Text.uncons (getLongName longName))
```
Gabriel439: `maybe mempty f = Data.Foldable.foldMap f`

src/Turtle/Options.hs
```diff
@@ -0,0 +1,100 @@
...
+   $ Opts.metavar (Text.unpack (getParameterName paramName))
+  <> optionalOr (Opts.help . Text.unpack . getHelpMessage) helpMessage
+
+newtype ParameterRead a = ParameterRead (ReaderT String Maybe a)
+    deriving (Functor, Applicative, Monad)
+
+parameterRead :: (Text -> Maybe a) -> ParameterRead a
+parameterRead f = ParameterRead (ReaderT (f . Text.pack))
+
+pAuto :: Read a => ParameterRead a
```
Gabriel439: Could you change the naming convention from `pAuto`/`pText`/`pInteger`/`pInt`/`pDouble` to `autoArg`/`textArg`/`integerArg`/`intArg`/`doubleArg`?  I try to avoid one-character prefixes or suffixes when naming things

int-index: Then it should be `autoParam`/`textParam`/etc. or `paramAuto`/`paramText`, because I use the term `Parameter`, not `Argument`. Also I prefer prefix `param` instead of suffix `param` because that way the type name starts with an uppercase letter.

Gabriel439:
I would suggest one slight restructure to simplify things further.  You could provide these types instead:

```haskell
textArg :: ParameterName -> Optional HelpMessage -> Parser Text
intArg :: ParameterName -> Optional HelpMessage -> Parser Int
doubleArg :: Parametername -> Optional HelpMessage -> Parser double
...

parameter :: (Text -> Maybe a) -> ParameterName -> Optional HelpMessage -> Parser a
```

The reason I suggest this is that it removes the intermediate `ParameterRead` type and users need one less source token for each option they add:

```haskell
options = Options
    <$> textArg    "NAME" "Your first name"
    <*> integerArg "AGE"  "Your current age"
    <*> switch "excited" "Are you excited?"
```

int-index:
Done. I agree that it's cleaner.

src/Turtle/Options.hs
```diff
@@ -0,0 +1,87 @@
...
+   $ Opts.metavar (Text.unpack (getParameterName paramName))
+  <> foldMap (Opts.help . Text.unpack . getHelpMessage) helpMessage
+
+parameterAuto :: Read a => ParameterName -> Optional HelpMessage -> Parser a
+parameterAuto = parameter (readMaybe . Text.unpack)
+
+parameterText :: ParameterName -> Optional HelpMessage -> Parser Text
+parameterText = parameter Just
+
+parameterInteger :: ParameterName -> Optional HelpMessage -> Parser Integer
```
Gabriel439: You can generalize the type to:

```haskell
parameterNumber :: Num n => ParameterName -> Optional HelpMessage -> Parser n
parameterNumber = fmap fromInteger (parameterAuto :: Parser Integer)
```

... and then you can use this to parse both `Int`s and `Integer`s

src/Turtle/Options.hs
```diff
@@ -0,0 +1,87 @@
...
+    , LongName(..)
+    , HelpMessage(..)
+    , options
+    , switch
+    , parameter
+    , parameterAuto
+    , parameterText
+    , parameterInteger
+    , parameterInt
+    , parameterDouble
```
Gabriel439: I suggest `s/parameter/arg/` for all of the above names, including `parameter`

src/Turtle/Options.hs
int-index: That would lead to confusion when `n` gets instantiated to `Double`, but `1.5` wouldn't parse.

src/Turtle/Options.hs
int-index: What about `ParameterName`? Should it be renamed to `ArgumentName`?

Also I'm not sure about `argAuto`. Perhaps `argRead` would be more descriptive because we're using the `Read` instance.

Gabriel439: Yeah, `ArgumentName` would be more consistent

I also like the idea of using `argRead`

src/Turtle/Options.hs
Gabriel439: You could change the constraint to `Integral` so that it would reject `Double` and other non-integral types

Gabriel439:
Awesome!  Thank you very much for contributing this!

I will add the documentation

Gabriel439:
I was just writing up the tutorial example and I realized that this actually creates positional arguments instead of flag-based arguments (i.e. "--name").  Would it be okay with you if I changed this so that if the user writes:

```
argText "name" "Your name"
```

... it creates a `--name NAME` flag?  The metavar would just be the same as the flag name except automatically upper-cased.

In other words, I would be replacing `Opts.argument` with `Opts.option`

int-index:
No, I don't want to get rid of positional arguments.

Let's extend current solution and have both positional arguments and flag-based options. `opt` similar to `arg`, `optText` similar to `argText`, etc.

PR #59, opened 2015-06-13
Fix inshell example in slides

PierreR:
As a note, this line is quite slow in ghci (~5s)


Gabriel439:
Yeah, the slowness seems to be due to the `process` library.  I'm still not sure why.

PR #79, opened 2015-06-25
Rename `du` to `size`

xged:
Originally `du` returned bytes instead of kilobytes, which I think is more useful. However *Unix du* returns kilobytes. So name change is appropriate.

Alternative option would be to keep name *du*, but change return value into kilobytes.

P.S. I do not know what is up with that  `Renamed 80 ./slides`, my commit msg was different (I am *newb* sorry)

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -123,7 +123,7 @@ module Turtle.Prelude (
     , rm
     , rmdir
     , rmtree
-    , du
+    , size
     , testfile
     , testdir
     , date
@@ -557,9 +557,9 @@ rmdir path = liftIO (Filesystem.removeDirectory path)
 rmtree :: MonadIO io => FilePath -> io ()
 rmtree path = liftIO (Filesystem.removeTree path)
 
--- | Get the size of a file or a directory in kilobytes
-du :: MonadIO io => FilePath -> io Integer
-du path = liftIO (Filesystem.getSize path)
+-- | Get the size of a file or a directory in bytes
+size :: MonadIO io => FilePath -> io Integer
+size path = liftIO (Filesystem.getSize path)
 
 -- | Check if a file exists
 testfile :: MonadIO io => FilePath -> io Bool


Gabriel439:
Another solution would be to change `du` to return an abstract `Size` type that you can extract sizes in different units.  In other words:

```haskell
du :: MonadIO io => FilePath -> IO Size

bytes :: Size -> Integer
kilobytes :: Size -> Integer
megabytes :: Size -> Integer
```

The advantage of that is that we now enforce the correct behavior in types instead of documentation.  The user can't accidentally interpret `Size` as the wrong size since they have to explicitly supply the units by using `bytes`/`kilobytes`/`megabytes`/etc.

xged:
I like it. The *Haskell* way. Another benefit is that you do not need to decide on default specific type then there is no obvious one.

3noch:
At the very least, you'd need to name it `fileSize` since `size` is far too generic.

Gabriel439:
I would prefer to keep the name to `du` to resemble Unix nomenclature

xged:
Fair enough

Talking about generic names there is `need` from *Turtle.Prelude* :)

3noch:
@xged At least need is not ambiguous. There are many types of sizes, not all of them are for files. Hard-drives can have a size, for example.

@Gabriel439 `du` is not as meaningful on Windows, but I suppose you must favor something or create a "normal" all of your own.

Gabriel439:
Turtle tries to stick with Unix naming conventions, even though it is portable to Windows.  The target audience is people who would typically write these scripts in Bash

xged:
@3noch Getting the size of a file or a directory is the primary use of the term *size*. That is not the case for the term *need*, let alone it is hard to google. Hard-drive is associated with the term *space* as well.

3noch:
@Gabriel439 Understood. That's a perfectly fine approach, but I will admit that Window's scripting is actually in more dire straights than Unix! :wink:

@xged Well `du` doesn't suffer from such debates anyway.

Gabriel439:
I will go ahead and close this mainly because I would like to keep the Unix names as much as possible.

PR #83, opened 2015-06-28
Additional changes to options parsing

Gabriel439:
@int-index: Could you take a look at these changes?  The main differences are:

* I generalized `argDouble` to `argFractional`
* I merged `argInt`/`argInteger`/`argIntegral` into `argIntegral`
* I changed the argument parsing to use flag-based options instead of positional arguments
* I removed short flag names to avoid clashes (it seems like these clashes would be frequent)

### DIFF
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -9,8 +9,8 @@
 -- > import Turtle
 -- >
 -- > parser :: Parser (Text, Int)
--- > parser = (,) <$> argText     "name" "Your first name"
--- >              <*> argIntegral "age"  "Your current age"
+-- > parser = (,) <$> optText     "name" "Your first name"
+-- >              <*> optIntegral "age"  "Your current age"
 -- >
 -- > main = do
 -- >     (name, age) <- options "Greeting script" parser
@@ -24,7 +24,7 @@
 -- > $ ./options --help
 -- > Greeting script
 -- >
--- > Usage: options --name NAME --age AGE
+-- > Usage: options (-n|--name NAME) (-a|--age AGE)
 -- >
 -- > Available options:
 -- >  -h,--help                Show this help text
@@ -38,8 +38,15 @@ module Turtle.Options
     , Description
     , HelpMessage
 
-      -- * Build parsers
+      -- * Flag-based option parsers
     , switch
+    , optText
+    , optIntegral
+    , optFractional
+    , optRead
+    , opt
+
+    -- * Positional argument parsers
     , argText
     , argIntegral
     , argFractional
@@ -73,9 +80,9 @@ options desc parser = liftIO
 
 {-| The name of a command-line argument
 
-    This is used to infer the long name and metavariable for the command line
-    flag.  For example, an `ArgName` of @\"name\"@ will create a @--name@ flag
-    with a @NAME@ metavariable
+    This is used to infer the long name, short name, and metavariable for the
+    command line flag.  For example, an `ArgName` of @\"name\"@ will create a
+    @--name@ flag with a @NAME@ metavariable and a short name of @-n@
 -}
 newtype ArgName = ArgName { getArgName :: Text }
     deriving (IsString)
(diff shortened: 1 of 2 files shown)


int-index:
1. Ok.
2. Ok.
3. Not ok. Firstly, it's inconsistent with `optparse-applicative` terminology (arguments are positional, options are keyword-based). Second, I want to have positional arguments. Let's have both. I can implement it.
4. Not sure. User should have simple ways to specify short name as:
    * specific letter
    * autogenerated from long name
    * no short name

int-index:
Perhaps `data ShortName = ShortName Char | DerivedShortName | NoShortName`?

Gabriel439:
Alright, then I'll add `argXXX` and `optXXX` versions of each option parser.  `argXXX` will be a positional argument and `optXXX` will be a flag-based option.

I'll go back to auto-generating short-names from the first character of long names, then.  My reasoning is that even in the case of conflicts it's still better than not having it at all.

int-index:
When I added auto-generation of short names I realized that conflicts are possible. My reasoning was that for small scripts it's easy to find option names that don't have a conflict, and when the amount of options grows really big, one can resort to full-featured `optparse-applicative`.

Gabriel439:
Alright, now there are both positional and flag-based versions of each parser

Gabriel439:
Yeah, that makes sense.  The users can always use the underlying API if this API is too restrictive for them.

PierreR:
I am a bit disappointed by the solution for `point 4`. I don't think it is user friendly enough to force "auto-generation of short name " and then say "well in case of conflict, go and use opt-applicative instead".

I believe it would make a lot of sense to have the opportunity to specify the short name with an optional arg. If the user does not specify the short name, then it is automatically auto-generated.

Gabriel439:
I'm still thinking of removing the short name to avoid conflicting short names.  My reasoning is that if the user wants something concise they can always use the positional arguments feature.

The other alternative is making the short name a mandatory argument instead of inferring it from the long name.

int-index:
@Gabriel439 What's your opinion on using something like `data ShortName = ShortName Char | DerivedShortName | NoShortName`?

Gabriel439:
I think it would be shorter and simpler to require the user to provide the short name.  It only adds four more characters, which is shorter than all the alternatives (other than removing it entirely)

int-index:
Why I'd prefer the `ShortName` data type:
1. Self-documenting constructor names
2. Easy to use
3. All three use cases satisfied (no short name, specific short name, derived short name)
4. No conflicting short names
5. Crippling functionality to save a few characters doesn't sound right to me

Gabriel439:
Actually, I think there's a good middle ground that offers all three alternatives while not being too verbose.  First off, I could make the short name `Optional`, so that all the option parsers have this type:

```haskell
optXXX :: ArgName -> Optional Char -> Optional HelpMessage -> Parser XXX
```

... and then add this `derived` combinator function:

```haskell
-- Maybe choose a better name other than `derived`
derived :: (ArgName -> Optional Char -> r) -> (ArgName -> r)
derived f argName = case Text.uncons (getArgName argName) of
    Nothing -> f argName empty
    Just (c, _) -> f argName c
```

So then the three options available to the user would be:

```haskell
-- Explicit short name
optXXX "long" (pure 'l') "help"

-- No short name
optXXX "long" empty "help"

-- Derived short name
derived optXXX "long" "help"
```

int-index:
@Gabriel439 

In #87 I argue that having `Char` for short names but dedicated types for everything else (`HelpMessage`, `ArgName`) is inconsistent.

Having a combinator like `derived` defeats your idea about having types as simple as possible. Regarding the verbosity:
1. `mempty` is almost as verbose as `empty`
2. `derived optxxx "long"` is as verbose as `optxxx "long" derive` where `derive = DeriveShortName`.
3. `pure 'l'` is shorter than `ShortName 'l'`, but I'm not sure if saving 5 characters is worth making the type signature less obvious.

After all, we can use abbreviations: `sn = ShortName`, `nsn = NoShortName`, `dsn = DeriveShortName`.

Gabriel439:
I still feel it's a lot of complexity to support the three options.  If it's a matter of making the type self-describing we can always use a type synonym:

```haskell
type ShortName = Char
```

... and still keep it mandatory.  I would really like to keep this as simple as possible.  I don't think most people will mind being asked to provide the short name explicitly.

int-index:
In that case let's have `type ShortName = Char`.

Gabriel439:
Alright.  By the way, thanks for taking the time to discuss this.  I really appreciate it!

PR #87, opened 2015-07-04
Updated options module

Gabriel439:
Changed arg/opt parsing to return monomorphic values and made short name
argument mandatory

### DIFF
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -35,6 +35,7 @@ module Turtle.Options
     ( -- * Types
       Parser
     , ArgName
+    , ShortName
     , Description
     , HelpMessage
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
@int-index : Please take a look at this if you have time

int-index:
The situation with short names is something I'm not content with. Here are the options we have:
1. Remove the short name functionality. Problem: the user script becomes clunky to use.
2. Require short names. Problem: mental burden on the script author to think of a short name for every option.
3. Derive short names from long names. Problem: conflicting short names.
4. A data type like `data ShortName = ShortName Char | DerivedShortName | NoShortName`. Problem: a few extra characters to type when writing a script.

I personally would prefer (4). Typing extra characters is something I can live with. The other problems seem more serious.

int-index:
Another argument in favor of a `ShortName` data type is that both type signatures and term-level code become self-documenting. It's more consistent with other design choices (having separate `ArgName` and `HelpMessage` but no `ShortName` is strange).

PR #90, opened 2015-07-20
Fix `find` example in documentation.

menelaos:


Gabriel439:
Thank you for catching that!

PR #99, opened 2015-08-18
`sed` now errors if passed a pattern matching the empty string

### LINKED ISSUE #96: Cause `sed` to error instead of looping when given a pattern which matches the empty string.
I feel like a call to `error` or `die` is a more appropriate solution than a warning in the documentation. 

(Apparently, unix `sed` inserts the substitution at every possible point in the string, which is not a behavior I think we'd want to imitate)

rpglover64:
Fixes #96

Unfortunately, this does not completely fix the problem. `match` is not sufficiently lazy, so `matchesEmpty` diverges if called with e.g. `many ""`.

A more complete solution would either change `match` to be sufficiently lazy or (even better) to keep track of which patterns match the empty string as part of the `Pattern` type and either error or short circuit in cases like `many ""`.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -958,11 +958,15 @@ grep pattern s = do
 -}
 sed :: Pattern Text -> Shell Text -> Shell Text
 sed pattern s = do
+    when (matchesEmpty pattern) (die message)
     let pattern' = fmap Text.concat
             (many (pattern <|> fmap Text.singleton anyChar))
     txt    <- s
     txt':_ <- return (match pattern' txt)
     return txt'
+  where
+    message = "sed: the given pattern matches the empty string"
+    matchesEmpty = not . null . flip match ""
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
My only suggestion is to maybe move the definition of `matchesEmpty` to inside of `sed` since it seems like a very specialized function.

src/Turtle/Prelude.hs
```diff
@@ -958,6 +958,7 @@ grep pattern s = do
 -}
 sed :: Pattern Text -> Shell Text -> Shell Text
 sed pattern s = do
+    when (matchesEmpty pattern) (die "sed was passed a pattern which matches the empty string")
```
Gabriel439: Also, maybe change this error message to `"sed: the given pattern matches the empty string"`

rpglover64:
The changes you've suggested have been made.

Gabriel439:
Thank you!

PR #107, opened 2015-09-26
Add text to tutorial describing the Stack shebang line

massysett:
I already knew what Stack is and was excited to see it can be used
in this manner, but I had to spend several minutes searching the web
to figure out what was going on with the second line in these
scripts.

(I closed an earlier pull request so I could change the Stack URL to a better one.)


Gabriel439:
Thanks!  It's a great idea to clarify how it works

PR #110, opened 2015-10-07
Make `mv` work accross filesystems

### LINKED ISSUE #37: mv only works within the same filesystem
mv uses Filesystem.rename which eventually calls rename(2). This will not work when src and dst are not on the same filesystem. The unix commando mv will do the right thing in these cases.

paulcapron:
Fixes #37 at the price of depending on GHC.IO.Exception.
Please see the commit message for (much!) more information.

Before:

    *Turtle> :set -XOverloadedStrings
    *Turtle> touch "/tmp/somefile" -- Here /tmp is on another filesystem (tmpfs)
    *Turtle> mv "/tmp/somefile" "./somefile"
    *** Exception: /tmp/somefile: rename: unsupported operation (Invalid cross-device link)

After:

    *Turtle> mv "/tmp/somefile" "./somefile"
    *Turtle> testfile "./somefile"
    True  -- Copy was done
    *Turtle> testfile "/tmp/somefile"
    False -- Removing was done

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -236,6 +236,7 @@ import qualified Data.Text.IO as Text
 import qualified Filesystem
 import Filesystem.Path.CurrentOS (FilePath, (</>))
 import qualified Filesystem.Path.CurrentOS as Filesystem
+import GHC.IO.Exception (IOErrorType(UnsupportedOperation))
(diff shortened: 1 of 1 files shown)


Gabriel439:
Fantastic work!  I'm definitely okay with depending on `GHC.IO.Exception`.

I also really love the commit message :)

PR #124, opened 2015-11-09
Make the option example more comprehensive

PierreR:
As a note, I don't quite see where "A task" would be used. Do you know ?

### DIFF
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -1,6 +1,10 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
--- | Example usage of this module:
+-- | Comprehensive example involving:
+-- |   - one subcommand (task)
+-- |   - one positional argument (TASK e.g Haskell)
+-- |   - two flag-based options (--name, --age):
+-- | (do not hesitate to go with a much simpler usage by using only flag-based options for instance)
 --
 -- > -- options.hs
 -- >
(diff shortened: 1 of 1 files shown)


PierreR:
Maybe col width up to 90 is a concern ? If so let me know.

Gabriel439:
Actually, I was thinking of doing two small examples instead of one large example.  The second example I had in mind for sub commands was a (fake) volume control application.  In other words, the user could write:

```haskell
$ ./volume up 10
$ ./volume down 20
```

That would illustrate subcommands and positional arguments.

PierreR:
Alright I can leave the first example as is and have a go writing the second as you suggest if you don't mind. There is still the mystery about the subcommand second argument

PierreR:
Maybe we should just make a link to the `Command line options section` on the tutorial ? And add the new example there ? I have myself overlooked the tutorial because of this example ...

PierreR:
Something that I find missing from the tutorial is how to provide some default for optional option (like the classic -p /--port flag)
Is there a little bug with subcommand that never shows its help text ?

PierreR:
Something else that does not seem to work is:
./options task --help

Gabriel439:
So I figured out why subcommands were not displaying `--help` output and fixed it in 4fd2edc0e0f70c4a90be2e2dc9e9e2e2b6a9b32a.  I'm still working on adding a subcommand example to the tutorial and linking to the tutorial from the `Turtle.Options` module.

Gabriel439:
Alright, I added a `subcommand` example to the tutorial and linked to the tutorial from the `Turtle.Options` documentation so I'll mark this as fixed

PierreR:
Thanks. Just one last question. Do you think the Option module could be packaged separately in the future ? I have found it much more easy to use and understand (even compared to `optparse-simple`). (same feeling with the `Format` module). I might give it a go and replace `optparse-applicative` with it by just copying the module and see how it fits in in a more complicated setting.

Gabriel439:
Yes, definitely!  I would also like to see `Turtle.Pattern` and `Turtle.Shell` packaged separately, too.  I'm just waiting for the library to stabilize before I start factoring things out.

Issue #113, opened 2015-10-21
inproc hangs if input stream terminates with error


stschiff:
Consider this little script 

    #!/usr/bin/env runghc
    {-# LANGUAGE OverloadedStrings #-}
    
    import Turtle
    
    main = do
        let stream1 = inproc "fdsdfd" [] empty
            stream2 = inproc "cat" [] stream1
        stdout stream2

This hangs immediately, whether compiled or interpreted.
Now consider putting `stdout stream1` in the last line, instead of `stdout stream2`. Now it correctly exits with an error:
    
    fdsdfd: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

But the first case, as written in the snippet hangs, and this is very problematic behaviour. This just hit me in the middle of a rather long-running script, and it took me a while to figure out that it meant that some input stream had an error. 

I use Turtle-1.2.2

Gabriel439:
Okay, so I managed to get this working when compiled by adding:

```diff
@@ -476,6 +480,7 @@ stream p s = do
             close hIn

     a <- using (fork feedIn)
+    liftIO (link a)
     inhandle hOut <|> (liftIO (Process.waitForProcess ph *> wait a) *> empty)

 -- | Print to @stdout@
```

... but I still hit the issue when interpreting the script.  I'm still investigating.

Gabriel439:
Oh, never mind, I misunderstood how `link` works.  It only checks once to see if the forked thread has thrown an exception at the time you call `link`; it doesn't keep checking continuously.  That explains why it gives different behavior when compiled vs. interpreted.  Okay, I think I know how to fix this, then.

Gabriel439:
Hmmm.  This is trickier than I thought.  I will need more time to figure out what is going on.

mitchellwrosen:
I think the `link`/`link2` docs are disingenuous. They are nothing like Erlang's `link` as you discovered. Any updates on this @Gabriel439?

Gabriel439:
I figured it out.  The subprocess runners weren't closing the input handle correctly if the input stream threw an exception.  See the diff of the fix to see what I mean.

mitchellwrosen:
Ah, intersting. But what if the input-feeding thread receives an async exception before doing anything, as marked with a `(*)` below? I think instead of e.g.

```haskell
bracket open (\(hIn, ph) -> close hIn >> Process.terminateProcess ph) (\(hIn, ph) -> do
    let feedIn =
[...]

PR #127, opened 2015-11-16
Further improve exception propagation for subprocess runners #113

mitchellwrosen:


Gabriel439:
Thanks! :)

mitchellwrosen:
Sorry, build failing on 7.6 _and_ I think I owe you another pull request regardless... `forkIOWithUnmask` does not automatically mask exceptions for you - child threads just inherit the masked state as always. So, this fix wasn't quite correct.

Gabriel439:
No problem

PR #128, opened 2015-11-16
Mask async exceptions before spawning subprocess runners

mitchellwrosen:
Continuation of #127 

Unfortunately I don't have a good way of getting this compiling on ghc 7.6 right now. `stack setup` says:

```
No information found for ghc-7.6.
Supported versions for OS key 'macosx': GhcVersion 7.8.4, GhcVersion 7.10.1, GhcVersion 7.10.2
```

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -221,7 +221,7 @@ import Control.Concurrent (threadDelay)
 import Control.Concurrent.Async (Async, withAsync, withAsyncWithUnmask, wait, concurrently)
 import Control.Concurrent.MVar (newMVar, modifyMVar_)
 import qualified Control.Concurrent.STM as STM
-import Control.Exception (bracket, finally, throwIO)
+import Control.Exception (bracket, finally, mask_, throwIO)
 import Control.Foldl (Fold, FoldM(..), genericLength, handles, list, premap)
 import qualified Control.Foldl.Text
 import Control.Monad (liftM, msum, when, unless)
@@ -373,7 +373,7 @@ system p s = liftIO (do
                     txt <- s
                     liftIO (Text.hPutStrLn hIn txt) ) )
                 `finally` close hIn
-        withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* wait a) ) )
+        mask_ (withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* wait a) ) ) )
 
 systemStrict
     :: MonadIO io
@@ -410,7 +410,7 @@ systemStrict p s = liftIO (do
                 `finally` close hIn
 
         concurrently
-            (withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* wait a))
+            (mask_ (withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* wait a)))
             (Text.hGetContents hOut) ) )
 
 {-| Run a command using @execvp@, streaming @stdout@ as lines of `Text`
@@ -477,7 +477,7 @@ stream p s = do
                 liftIO (Text.hPutStrLn hIn txt) ) )
             `finally` close hIn
 
-    a <- using (managed (withAsyncWithUnmask feedIn))
+    a <- using (managed (mask_ . withAsyncWithUnmask feedIn))
     inhandle hOut <|> (liftIO (Process.waitForProcess ph *> wait a) *> empty)
 
 -- | Print to @stdout@
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE CPP                        #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RankNTypes                 #-}
 
 -- | This module provides a large suite of utilities that resemble Unix
 --  utilities.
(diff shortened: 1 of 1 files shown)


Gabriel439:
What's the purpose of the `mask_`?  Also, I think you need to give a type signature to `feedIn` to prove to the compiler that it accepts a higher-rank argument

mitchellwrosen:
Without `mask_`, an async exception thrown to the `feedIn` thread before it calls `restore` will bring the thread down because its masking state is `Unmasked`. `withAsyncWithUnmasked` doesn't automatically mask async exceptions for you, it only provides you with an unmasking function.

Gabriel439:
I believe the type signature to `feedIn` needs to be:

```haskell
feedIn :: (forall c . IO c -> IO c) - >...
```

mitchellwrosen:
Huh, I wonder why it compiles on 7.8+, then.

Gabriel439:
Yeah, I'm not sure either.  Maybe `ghc-7.8+` does a better job of inferring higher-rank types

PierreR:
I kind of understand why we need to do the `mask_/withAsyncWithUnmask` combo but I still find it surprisingly hard to read (and get right) specially coming from `async` (an high level library).

mitchellwrosen:
I agree, and I think this deserves a second look. The problem is that `feedIn`, run in a separate thread, has its own teardown procedure. I don't think `async` offers a cleaner way to achieve this, other than the one I used (`mask_ . withAsyncWithUnmask`).

But what about just moving the exception handling to the main thread?

```haskell
bracket open (\(hIn, ph) -> close hIn >> Process.terminateProcess ph) (\(hIn, ph) -> do
    let feedIn :: IO ()
        feedIn =
            sh (do
                txt <- s
                liftIO (Text.hPutStrLn hIn txt) )
        withAsync (feedIn `finally` close hIn) (\a -> liftIO (Process.waitForProcess ph) <* wait a) ) )
```

Gabriel439:
What you just wrote is identical to what I originally had :)

mitchellwrosen:
So it is :) I think then I was wrong about the original code having a subtle bug. I basically spent half a day re-inventing part of `async`.

Gabriel439:
I can revert the `mask_ . withAsyncUnmask` changes then if they aren't necessary

mitchellwrosen:
:+1: from the guy wearing the dunce cap

Gabriel439:
:)  Don't worry, I was just as confused myself.  Reasoning about asynchronous exception safety is very difficult

mitchellwrosen:
Okay, I've thought about this for another half day and I'm convinced I was
right the first time around for all the same reasons. The original code
does have a subtle bug which was fixed by `mask_ . withAsynsWithUnmask`,
and then I somehow convinced myself there was a "cleaner" way.

Anyway, the bug will probably not bite anyone, but in theory it is possible.

Gabriel439:
Yeah, from looking at the code for `withAsync` more closely, it internally uses `withAsyncUsing` which does provide an opportunity for an exception to leak in so I think we can leave the code as you wrote it.

mitchellwrosen:
Where do you see that? Not sure if you saw my EDIT above but I flip flopped yet again. I believe your initial implementation is correct.

mitchellwrosen:
Here's a common way for an exception to leak:

```
        async exception can strike here
        |
        v
forkIO ( foo `finally` bar)
```

and here's the Haskell solution:

```
                      exception cannot prevent `bar` from running
                      |
                      v
mask_ (forkIO (restore foo `finally` bar))
```

which is exactly what `withAsync` does for you.

Gabriel439:
Yeah, I saw your edit, but looking at the code, it is essentially doing this:

```haskell
mask (\restore -> do
    ...
    restore (feedIn `finally` closeIn)
    ... )
```

... and that provides an opportunity for an exception to leak in right after the `restore` begins but before the `feedIn` begins.  Or to make this more explicit:

```haskell
mask (\restore -> do
    ...
    restore (do
        -- <= Exception can leak in right here
        feedIn `finally` closeIn )
    ... )
```

mitchellwrosen:
If an exception could strike there, I don't think I could sleep at night. `restore` shouldn't allow exceptions to strike _before_ the provided argument.

Gabriel439:
My understanding was that this:

```haskell
restore (feedIn `finally` closeIn`)
```

... is not exception safe, but this:

```haskell
(restore feedIn) `finally` closeIn
```

... is exception-safe.  However, it's been a while since I last tested this so my memory might be faulty.  Most of what I learned about exception safety was a long time ago when working on `pipes-safe`.

mitchellwrosen:
I think in general you're correct, but `finally` itself masks exceptions!

Gabriel439:
Right, but `finally` hasn't had a chance to run yet!  The exception strike befores `finally` begins.

mitchellwrosen:
So you're saying the exception can strike here?

```
         |
         v
restore ( mask $ \restore2 -> do
    restore2 feedIn `onException` closeIn
    closeIn)
```

Gabriel439:
Yes

mitchellwrosen:
That's true if the act of unmasking exceptions and then running the provided action is not atomic. I can't really tell from reading over the Control.Exception docs for the umpteenth time. Is it in there somewhere?

Gabriel439:
I crafted a small example to prove that an exception can leak in right at the beginning of a `restore`:

```haskell
import Control.Concurrent
import Control.Exception

prints1 =
    uninterruptibleMask (\restore -> do
        me <- myThreadId
        forkIO (killThread me)
        restore (threadDelay 1000000 `finally` print 2) )
    `finally` print 1

prints1And2 =
    uninterruptibleMask (\restore -> do
        me <- myThreadId
        forkIO (killThread me)
        restore (threadDelay 1000000) `finally` print 2 )
    `finally` print 1

main = prints1
```

mitchellwrosen:
Wow, bravo. Ok, then turns out I was right all along! Nailed it.

Gabriel439:
Yeah! :)

Now you know what I mean when I say that it's hard to reason about asynchronous exception safety.  They can literally strike **anywhere**

PR #144, opened 2016-03-06
Change `rmtree` on Unix-symlinks. Fixes #143

### LINKED ISSUE #143: `rmtree` cannot remove files or symlinks.
The [documentation of `rmtree`](https://github.com/Gabriel439/Haskell-Turtle-Library/blob/master/src/Turtle/Prelude.hs#L848-L853) seems to be misleading. It says that `rmtree` would be equivalent to the shell command `rm -r`.

However, `rmtree` cannot delete files. Instead it produces the following error when called as `rmtree "file"`:

    *** Exception: file: listDirectory: inappropriate type (Not a directory)

And, `rmtree` cannot delete symlinks. Instead it produces the following error when called as `rmtree "symlink"`:

    *** Exception: symlink: removeDirectory: inappropriate type (Not a directory)

The behavior of `rm -r` is quite different (at least on Linux).

1. It removes files (i.e. not only directories)
2. If the given path is a symlink, then it will remove the symlink, but not the contents of the directory that the symlink points to.

aherrmann:
Before `rmtree` would call `Filesystem.isDirectory` on the symlink, which
resolves the symlink _before_ testing whether it is a directory.
If the symlink refers to the directory, then it would delete all the contents
of that directory and afterwards fail to remove the symlink itself.

Now it behaves like `rm -r` on unix systems. When called on a symlink it just
deletes that symlink irregardles of what it points to.

Unfortunately, I don't know how to deal with symlinks on Windows systems.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -281,7 +281,14 @@ import qualified System.Process as Process
 #ifdef mingw32_HOST_OS
 import qualified System.Win32 as Win32
 #else
-import System.Posix (openDirStream, readDirStream, closeDirStream, touchFile)
+import System.Posix (
+    openDirStream,
+    readDirStream,
+    closeDirStream,
+    touchFile,
+    getSymbolicLinkStatus,
+    isDirectory,
+    isSymbolicLink )
 #endif
 import Prelude hiding (FilePath)
 
@@ -865,6 +872,7 @@ rmdir path = liftIO (Filesystem.removeDirectory path)
 rmtree :: MonadIO io => FilePath -> io ()
 rmtree path0 = liftIO (sh (loop path0))
   where
+#ifdef mingw32_HOST_OS
     loop path = do
         isDir <- testdir path
         if isDir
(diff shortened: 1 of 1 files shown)


Gabriel439:
Thanks for fixing that!  I can work on getting the Windows support for this working correctly

aherrmann:
Thanks for merging!

OT: Are there any plans to also be able to create symlinks with turtle (other than by calling `ln -s`)? [This Stack Overflow discussion](http://stackoverflow.com/questions/35369470/what-is-the-command-to-create-a-soft-link-with-haskell-turtle) indicates that it would be difficult to get right on Windows.

Gabriel439:
I haven't been actively working on symlinks recently but the last time I took a stab at working with them (via the `unix-compat` library) I ran into some issues (although I forgot what they were)

aherrmann:
I just had a look at that package. It's probably that [they're unsupported](https://github.com/jystic/unix-compat/blob/master/src/System/PosixCompat/Files.hsc#L374-L375).

PR #146, opened 2016-03-15
relax process upper bound

joehillen:
I need process-1.4.2.0 for the [NoStream](https://hackage.haskell.org/package/process-1.4.2.0/docs/System-Process.html#t:StdStream). I tested things and it works fine.


Gabriel439:
Yeah, I didn't realize `process` had so many releases since I last checked.  Thanks for the update

PR #149, opened 2016-03-25
Add a grep specific to FilePath

yogsototh:
Don't you think it would be a nice idea to propose this function directly?
And may be to provide an example in the tutorial?

Typically in shell when you are doing `**/*` it won't match files and directories which start by a `.`:

~~~ {.haskell}
allFiles <- fpgrep (invert (prefix "./.")) (find (has ".hs") ".")
~~~

Best.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -1260,6 +1260,14 @@ grep pattern s = do
     _:_ <- return (match pattern txt)
     return txt
 
+-- | Keep all files that match the given `Pattern`
+fpgrep :: Pattern a -> Shell FilePath -> Shell FilePath
+fpgrep pattern s = do
+  fpath <- s
+  _:_ <- return (match pattern (either id id (toText fpath)))
+  return fpath
+
+
 {-| Replace all occurrences of a `Pattern` with its `Text` result
 
     `sed` performs substitution on a line-by-line basis, meaning that
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -173,6 +173,7 @@ module Turtle.Prelude (
     , lstree
     , cat
     , grep
+    , fpgrep
     , sed
     , inplace
     , find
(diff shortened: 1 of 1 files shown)


Gabriel439:
How about a more general utility like this:

```haskell
onFiles
    :: (Shell FilePath -> Shell FilePath)
    -> (Shell Text     -> Shell Text    )
```

Then you could not only use it with `grep`, but also with `sed` and `cache`

yogsototh:
@Gabriel439 I would love to have a more generic utility, but I fail to see how I could achieve my example use case with the `onFiles`. In the end, I need to have a `Shell FilePath`.

The closer I could get is

~~~ {.haskell}
onFiles :: (Text -> FilePath -> [b]) -> Shell FilePath -> Shell b
onFiles trans sa = do
  anA <- sa
  let bs = trans (either id id (toText anA)) anA
  select bs
~~~ 

From which I could do:

~~~ {.haskell}
fpgrep :: Pattern a -> Shell FilePath -> Shell FilePath
fpgrep pat = onFiles $ \ txtpath fpath = match pat txtpath >> return fpath
~~~

or for my end use case:

~~~ {.haskell}
allFiles <- onFiles removeStartingByDot  (find (has ".hs") ".")

removeStartingByDot :: Text -> FilePath -> [FilePath]
removeStartingByDot txtPath fpath = match pat txtpath >> return fpath
~~~

Of course I could also make something more general with:

~~~ {.haskell}
onS :: (a -> [b]) -> Shell a -> Shell b
onS trans sa = do
  anA <- sa
  select (trans anA)

fpgrep :: Pattern a -> Shell FilePath -> Shell FilePath
fpgrep pat =
  on (\fpath -> match pat (either id id (toText fpath)) >> return fpath)
~~~

But I have to _see_ the `either id id (toText fpath)` in my code. Which is exactly what I wanted to prevent and the first reason of this pull request.

Do you see a cleaner way?

Gabriel439:
You would use `onFiles` like this:

```haskell
allFiles <- onFiles (grep (invert (prefix "./."))) (find (has ".hs") ".")
```

The reason this works is that `grep pattern` has type `Shell Text -> Shell Text`, and when you wrap it with `onFiles` now it has type `Shell FilePath -> Shell FilePath`.

devonhollowood:
@yogsototh Are you still working on this? If not, I'll give it a shot.

yogsototh:
@devonhollowood not actively I'm afraid. Do as you wish!

Gabriel439:
Alright, so now we have an `onFiles` contributed by @devonhollowood in #180 so I'm going to mark this as fixed

PR #159, opened 2016-07-09
Two minor tweaks to make it work with ghci

erikd:


Gabriel439:
Could you explain the `__GLASGOW_HASKELL__` change?  Why does that work instead of `MIN_VERSION_base`?

erikd:
I'm not sure why `MIN_VERSION_base` is not available in ghci run with `cabal exec -- ghci ...` but changing to `__GLASGOW_HASKELL__` made it work. I found that solution at https://gist.github.com/nh2/c1aac83e9561dd78b8e3 .

Since the version of base is always initimately tied to the GHC version, this should be safe.

lspitzner:
because cabal has no idea that the source is using CPP. i think you can add `other-extensions: CPP` to the different components to fix.

erikd:
@lspitzner The `.ghci` file in the current directory sets `CPP`. I've also tried adding `-XCPP` to the ghci command line. I think the problem is that the `MIN_VERSION_base` macro is something that cabal generates and ghci loads the file directly. The `__GLASGOW_HASKELL__` defines are in GHC itself.

lspitzner:
@erikd Setting `-XCPP` is not the problem, that is already done by the LANGUAGE pragmas in each source file, where required. The problem is indeed the macro generation, which currently does not happen because `cabal` is not aware that it is supposed to. With my proposed addition to the `.cabal` file, `cabal repl` _should_ properly create the macros.

lspitzner:
my proposed fix does not make a pure `ghci` work, and probably neither a `cabal exec -- ghci`, because no macros are generated at all in the first place, and they are not forwarded to ghci in any appopriate fashion in the second case.

Imho fixing the `.cabal` and using `cabal repl` is the appropriate solution. Of course @Gabriel439 is free to include the current patch as well to support other workflows, but personally i don't see the need.

erikd:
Unfortunately, I have never been satisfied with `cabal repl` and sure enough, in this particular case does not work.

The problem is that `cabal repl` pulls in every file mentioned in the cabal file (which I consider a mis-feature), including `Turtle.Tutorial` which is documentation only (it contains no actual code) and that documentation causes other errors.

```
$ cabal repl src/Turtle.hs 
Package has never been configured. Configuring with default flags. If this
fails, please run configure manually.
Resolving dependencies...
Configuring turtle-1.2.8...
Warning: Ignoring 'src/Turtle.hs. The whole library will be built. (Support
for module and file targets has not been implemented yet.)
Preprocessing library turtle-1.2.8...
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help

src/Turtle/Tutorial.hs:132:4:
     warning: backslash and newline separated by space
     -- \ 
        ^
```

lspitzner:
but that error only occurs because you enabled CPP globally...

apart from that, `cabal repl` loads without problems for me without any changes to current HEAD.

And if you add `{-# LANGUAGE NoCPP #-}` on top of `Tutorial.hs`, cabal repl loads even with globally enabled `CPP`.

lspitzner:
(that it works fine of course counters my assumption that `other-extensions` is necessary. but perhaps that depends on cabal version used.)

lspitzner:
(but i have made bad experiences with `cabal repl` as well, so i certainly don't wish to dismiss your point in general.)

erikd:
Ok, so it might be possible to get `cabal repl` working in this particular instance, but in general I find `cabal repl` a pain in the neck (mainly because it attempts to loads everything in the cabal file and during development some things may be broken). I also like to keep the number of tools I use to a minimum. Having to use `cabal repl` here and `cabal exec -- ghci` elsewhere imposes extra cognitive load and development process friction that I'd like to avoid.

erikd:
I just realized that the GHCi shipped with ghc-8.0 finds the `MIN_VERSION_base`macros, but GHCi shipped with ghc-810 and earlier does not.

erikd:
@Gabriel439 ping?

These two patches are not needed for ghc-8.0, but with ghc 7.10 and earlier are required to load Turtle into ghci.

Gabriel439:
Sorry for the delay!  This seems fine, then

erikd:
Thanks @Gabriel439 !

Gabriel439:
@erikd Thank you, too! :)

PR #164, opened 2016-08-03
Add pushd

maoe:
This PR adds a new function `pushd`. It behaves like the `pushd` in bash but there's no `popd` equivalent. Instead it moves back to the original directory when the enclosing `Shell` monad is done. See the doctest for an example.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -153,6 +153,7 @@ module Turtle.Prelude (
     , mktempdir
     , fork
     , wait
+    , pushd
 
     -- * Shell
     , inproc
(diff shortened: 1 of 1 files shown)


maoe:
In case `popd` is necessary (I don't have a compelling use case though), this might make sense:

```diff
-pushd :: MonadManaged managed => FilePath -> managed ()
+pushd :: (MonadIO io, MonadManaged managed) => FilePath -> managed (io ())
 pushd path = do
     cwd <- pwd
     using (managed_ (bracket_ (cd path) (cd cwd)))
+    return $ cd cwd
+
+pushd_ :: MonadManaged managed => FilePath -> managed ()
+pushd_ path = do
+    (_popd :: IO ()) <- pushd path
+    return ()
```

I prefer the original implementation because of simplicity though.

Gabriel439:
Looks great!  Thank you for contributing this :)

PR #165, opened 2016-08-19
Fixed documentation syntax error.

GregorySchwartz:
"/" is a special character and it was making the dots...italic.


Gabriel439:
Thanks for the fix! :)

PR #166, opened 2016-08-22
PoC: Add json :: Shell Text -> Shell Value

maoe:
I came up with an idea to stream texts into JSON values. It's still a PoC as I do have a use case but I haven't actually tried it yet.

Some concerns:
* The extra dependencies might be a concern. Worth a separate package?
* May or may not want to use `Aeson.value'` instead

What do you think?

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -184,6 +184,7 @@ module Turtle.Prelude (
     , limit
     , limitWhile
     , cache
+    , json
 
     -- * Folds
     , countChars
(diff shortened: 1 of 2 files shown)


Gabriel439:
My rough rule of thumb for what should go into the library is whether there is some equivalent Unix utility.  Is there some Unix analog of this functionality?

maoe:
Well, that's a good question. I think the [`jq`](https://stedolan.github.io/jq/) command is somewhat similar in the sense that it parses texts as JSON objects. `jq` has a query language to look up or edit the parsed JSON objects. We can do similar things with `lens-aeson` for example.

Gabriel439:
So I think that `jq`-like functionality is desirable, but I would like to see some examples of how to translate various `jq` idioms to Haskell

maoe:
Okay, I'll translate the examples on `jq` website.

Issue #160, opened 2016-07-14
Use Doc rather than Text for options header


mitchellwrosen:
What do you think about 
```haskell
newtype Description = Description { getDescription :: Doc }
```
rather than
```haskell
newtype Description = Description { getDescription :: Text }
```
?

You still have an `IsString` instance, and moreover users can format headers to their liking. Currently, `Options.Applicative.Builder.header` just smashes all text fragments together, even they contain newlines and such.

Gabriel439:
Yeah, I think that's a great idea!

PR #170, opened 2016-09-03
Use Doc instead of Text for program description

mitchellwrosen:
#160


Gabriel439:
Awesome!  Looks great :)

Issue #38, opened 2015-04-07
Add stat :: FilePath -> IO FileStatus


maoe:
It would be nice if we had `stat :: FilePath -> IO FileStatus`.  It doesn't seem to be too difficult if we can use `unix` for POSIX systems and `unix-compat` for others but it incurs an extra dependency (only for non POSIX systems). What do you think?

Gabriel439:
I didn't even know `unix-compat` existed!  Thanks for pointing that out.

I will try to add this and several other functions from that package.

Gabriel439:
This was fixed by #171

PR #171, opened 2016-09-03
Add stat, lstat

mitchellwrosen:
#38

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -228,6 +228,22 @@ module Turtle.Prelude (
     , gibibytes
     , tebibytes
 
+    -- * File status
+    , PosixCompat.FileStatus
+    , stat
+    , lstat
+    , fileSize
+    , accessTime
+    , modificationTime
+    , statusChangeTime
+    , PosixCompat.isBlockDevice
+    , PosixCompat.isCharacterDevice
+    , PosixCompat.isNamedPipe
+    , PosixCompat.isRegularFile
+    , PosixCompat.isDirectory
+    , PosixCompat.isSymbolicLink
+    , PosixCompat.isSocket
+
     -- * Exceptions
     , ProcFailed(..)
     , ShellFailed(..)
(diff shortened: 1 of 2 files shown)


src/Turtle/Prelude.hs
```diff
@@ -228,6 +228,29 @@ module Turtle.Prelude (
...
+    , PosixCompat.deviceID
+    , PosixCompat.fileID
+    , PosixCompat.fileMode
+    , PosixCompat.linkCount
+    , PosixCompat.fileOwner
+    , PosixCompat.fileGroup
+    , PosixCompat.specialDeviceID
+    , PosixCompat.fileSize
+    , PosixCompat.accessTime
+    , PosixCompat.modificationTime
```
Gabriel439: For `stat`-related functions that return an `EpochTime` like `accessTime` and `modificationTime`, could we wrap them to return a `time` library type like `PosixTime`?

src/Turtle/Prelude.hs
```diff
@@ -228,6 +228,29 @@ module Turtle.Prelude (
...
+    , stat
+    , lstat
+    , PosixCompat.deviceID
+    , PosixCompat.fileID
+    , PosixCompat.fileMode
+    , PosixCompat.linkCount
+    , PosixCompat.fileOwner
+    , PosixCompat.fileGroup
+    , PosixCompat.specialDeviceID
+    , PosixCompat.fileSize
```
Gabriel439: Could we wrap this to return a `Size`?

src/Turtle/Prelude.hs
```diff
@@ -228,6 +228,29 @@ module Turtle.Prelude (
...
+    , PosixCompat.FileStatus
+    , stat
+    , lstat
+    , PosixCompat.deviceID
+    , PosixCompat.fileID
+    , PosixCompat.fileMode
+    , PosixCompat.linkCount
+    , PosixCompat.fileOwner
+    , PosixCompat.fileGroup
+    , PosixCompat.specialDeviceID
```
Gabriel439: For all the functions from `specialDeviceID` and up I'd suggest removing them until somebody needs them.  My reasoning is that it's not clear what more opinionated and idiomatic type to wrap them in, and we won't know the answer to that until somebody needs to use them.

Gabriel439: ... except of course the `stat` and `lstat` functions.  Keep those! :)

Gabriel439:
Other than those comments this looks good to me

mitchellwrosen:
Done and force-pushed

Gabriel439:
Awesome!  Thank you :)

Also, for future reference, you don't need to force push if you want a clean history.  Github provides me with the option to perform a squash merge that generates a single commit from your branch, no matter how messy it is.

mitchellwrosen:
Oh, I didn't know that, thanks!

PR #172, opened 2016-09-03
Add which

mitchellwrosen:
#7

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE CPP                        #-}
 {-# LANGUAGE DeriveDataTypeable         #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes                 #-}
@@ -136,6 +137,7 @@ module Turtle.Prelude (
     , touch
     , time
     , hostname
+    , which
     , sleep
     , exit
     , die
@@ -295,7 +297,8 @@ import System.Exit (ExitCode(..), exitWith)
 import System.IO (Handle, hClose)
 import qualified System.IO as IO
 import System.IO.Temp (withTempDirectory, withTempFile)
-import System.IO.Error (catchIOError, ioeGetErrorType)
+import System.IO.Error
+    (catchIOError, ioeGetErrorType, isPermissionError, isDoesNotExistError)
 import qualified System.PosixCompat as PosixCompat
 import qualified System.Process as Process
 #ifdef mingw32_HOST_OS
@@ -1094,6 +1097,32 @@ time io = do
 -- | Get the system's host name
 hostname :: MonadIO io => io Text
 hostname = liftIO (fmap Text.pack getHostName)
+
+-- | Show the full path of an executable file
+which :: MonadIO io => FilePath -> io (Maybe FilePath)
+which cmd = fold go Control.Foldl.head
+  where
+    go :: Shell FilePath
+    go =
+      need "PATH" >>= \case
+          Nothing -> empty
+          Just paths -> do
+            path <- select (Text.split (== ':') paths)
+            let path' = Filesystem.fromText path </> cmd
+            exists <- testfile path'
+            if exists
+                then do
+                    perms <- liftIO $
+                        getmod path'
+                          `catchIOError`
+                            \e ->
+                                if isPermissionError e || isDoesNotExistError e
+                                    then return Directory.emptyPermissions
+                                    else throwIO e
+                    if Directory.executable perms
+                        then return path'
+                        else empty
+                else empty
 
 {-| Sleep for the given duration
 
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -138,6 +138,7 @@ module Turtle.Prelude (
     , time
     , hostname
     , which
+    , whichAll
     , sleep
     , exit
     , die
@@ -261,7 +262,7 @@ import Control.Exception (Exception, bracket, bracket_, finally, mask_, throwIO)
 import Control.Foldl (Fold, FoldM(..), genericLength, handles, list, premap)
 import qualified Control.Foldl
 import qualified Control.Foldl.Text
-import Control.Monad (liftM, msum, when, unless)
+import Control.Monad (guard, liftM, msum, when, unless)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Managed (MonadManaged(..), managed, managed_, runManaged)
 #ifdef mingw32_HOST_OS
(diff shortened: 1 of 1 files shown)


Gabriel439:
Could you merge `master` again?  The build failure in CI is my fault and is an issue in `master` that I just fixed

src/Turtle/Prelude.hs
```diff
@@ -1090,6 +1093,32 @@ time io = do
...
+          Just paths -> do
+            path <- select (Text.split (== ':') paths)
+            let path' = Filesystem.fromText path </> cmd
+            exists <- testfile path'
+            if exists
+                then do
+                    perms <- liftIO $
+                        getmod path'
+                          `catchIOError`
+                            \e ->
```
Gabriel439: One minor suggestion to improve readability.  Perhaps factor out the handler as a named function, like this:

```haskell
...
then do
    let handler e =
            if isPermissionError e || isDoesNotExistError e
                then return Directory.emptyPermissions
                else throwIO e
    perms <- liftIO (getmod path' `catchIOError` handler)
    ...
```

The other reason that I like this trick is that you can add a type signature to `handler` to specify the type of the exception to catch (although it's not necessary in this example)

src/Turtle/Prelude.hs
```diff
@@ -1090,6 +1093,32 @@ time io = do
 hostname :: MonadIO io => io Text
 hostname = liftIO (fmap Text.pack getHostName)
 
+-- | Show the full path of an executable file
+which :: MonadIO io => FilePath -> io (Maybe FilePath)
+which cmd = fold go Control.Foldl.head
```
Gabriel439: I was wondering if it would be worth just returning `go` (i.e. change the result to `Shell FilePath`).  This would allow the user to retrieve all matching programs on the path.  For the common case they could wrap `which` in `fold which Control.Foldl.head`, but if they wanted even better error handling they could do something like:

```haskell
example = do
    paths <- fold go Control.Foldl.list
    case paths of
        []     -> ... -- Warn the user that no matches are found
        [path] -> ... -- Do something with the path
        _      -> ... -- Warn the user that there are multiple matches
```

I don't feel strongly about this, though

Gabriel439:
Other than those comments this looks good to me

src/Turtle/Prelude.hs
mitchellwrosen: That does look nicer, but it doesn't work when the action is a do block! How would you structure this?

```haskell
then do
    let handler e = ...
         action = do
             ....
             ....
    action `catcIOError` handler
```

(assuming that `let action = ...` isn't ideal)

src/Turtle/Prelude.hs
mitchellwrosen: I don't feel strongly about it either, but the list behavior is closer to `which --all`, if that's worth anything.

src/Turtle/Prelude.hs
```diff
@@ -1090,6 +1093,32 @@ time io = do
...
+  where
+    go :: Shell FilePath
+    go =
+      need "PATH" >>= \case
+          Nothing -> empty
+          Just paths -> do
+            path <- select (Text.split (== ':') paths)
+            let path' = Filesystem.fromText path </> cmd
+            exists <- testfile path'
+            if exists
```
int-index: I believe this `if` can be replaced by `guard exists`.

src/Turtle/Prelude.hs
```diff
@@ -1090,6 +1093,32 @@ time io = do
...
+                then do
+                    perms <- liftIO $
+                        getmod path'
+                          `catchIOError`
+                            \e ->
+                                if isPermissionError e || isDoesNotExistError e
+                                    then return Directory.emptyPermissions
+                                    else throwIO e
+                    if Directory.executable perms
```
int-index: Likewise, this `if` can be replaced by `guard (Directory.executable perms)`.

src/Turtle/Prelude.hs
int-index: Sound like a good argument in favor of the list behavior.

src/Turtle/Prelude.hs
Gabriel439: If I'm reading the  code correctly, the `catchIOError` only scopes over the `getMod`, which is not in a `do` block.  However, the answer to your question is that I would write it exactly the way you did, (i.e. a separate `let` binding for `action`)

src/Turtle/Prelude.hs
Gabriel439: Oh, I didn't know about `which --all`?  Then how about adding the `Shell` version as `whichAll` and the `head` version as `which`?

src/Turtle/Prelude.hs
mitchellwrosen: It is only scoping over the `getMod`, only because I couldn't make it look pretty the other way :)

No point in catching the error and returning empty permissions to inspect - could instead just catch the exception outside of the permission check and return empty there.

src/Turtle/Prelude.hs
mitchellwrosen: @int-index did you mean what Gabriel said? Or did you mean that `which` should return a Shell?

int-index: I meant `io [FilePath]` but I like @Gabriel439's proposal more.

mitchellwrosen:
Done

src/Turtle/Prelude.hs
```diff
@@ -1095,6 +1099,33 @@ time io = do
...
 hostname = liftIO (fmap Text.pack getHostName)
 
+-- | Show the full path of an executable file
+which :: MonadIO io => FilePath -> io (Maybe FilePath)
+which cmd = fold (whichAll cmd) Control.Foldl.head
+
+-- | Show all matching executables in PATH, not just the first
+whichAll :: FilePath -> Shell FilePath
+whichAll cmd =
+    need "PATH" >>= \case
```
int-index: `Shell` handles pattern failures as `empty`, so this line can be written as `Just paths <- need "PATH"`, removing another indentation level.

mitchellwrosen: Sweet. Moment.

mitchellwrosen: Done

src/Turtle/Prelude.hs
```diff
@@ -1095,6 +1098,31 @@ time io = do
...
+which cmd = fold (whichAll cmd) Control.Foldl.head
+
+-- | Show all matching executables in PATH, not just the first
+whichAll :: FilePath -> Shell FilePath
+whichAll cmd = do
+  Just paths <- need "PATH"
+  path <- select (Text.split (== ':') paths)
+  let path' = Filesystem.fromText path </> cmd
+
+  exists <- testfile path'
```
Gabriel439: You can go a step further and apply the same pattern match failure trick here:

```haskell
True <- testfile path'
```

Then you don't need the `guard`

mitchellwrosen: And the second `guard`?

```haskell
let handler :: IOError -> Permissions
     handler = ...

perms <- liftIO (getmod path' `catchIOError` handler)
guard (executable perms)
```
or
```haskell
let handler :: IOError -> Bool
     handler = ...

True <- liftIO (fmap executable (getmod path') `catchIOError` handler)
```

int-index: The second `guard` is more readable as is, in my opinion.

int-index:
The PR needs another rebase on `master` to fix the build failure on GHC 7.6.3 (which was due to `need` being unavailable).

int-index:
Have you tested that the function works correctly in all cases? If yes, it is good to merge.

mitchellwrosen:
To be honest I didn't do much testing, just poked at it with a stick in ghci. It's pretty simple, though: tokenize $PATH, see if there is an executable thing there. I'm sure it fails in weird corner cases, like if there's a directory in $PATH that is `-x` or something.

Is there any specific test you'd like to see?

Gabriel439:
Note that this is basically the same behavior as the Unix `which` so it should be well-behaved.  See: https://opensource.apple.com/source/shell_cmds/shell_cmds-118/which/which.c

Gabriel439:
Alright, this looks great to me!  I'll go ahead and squash and merge.  Thanks for the contribution! :)

PR #176, opened 2016-09-10
Add ability to capture (Exit Code, stdout, stderr)

### LINKED ISSUE #139: No way to capture stdout, stderr and exitcode
`inshellWithErr`/`inprocWithErr` are stdout and stderr; `procStrict`/`shellStrict` are stdout and exitCode.

It is not documented if the former throw some exception if the exitcode is not 0, but i would assume not (?). So maybe there is _some_ way.

Also, the param documentation of `inprocWithErr` and `inshellWithErr` does not mention stderr, which had me confused, because i searched for stderr on the haddock.

devonhollowood:
Add two new public functions: procStrictWithErr and shellStrictWithErr.
Implement them in terms of a new private function, systemStrictWithErr.

Fix #139

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -201,6 +201,8 @@ module Turtle.Prelude (
     , shells
     , procStrict
     , shellStrict
+    , procStrictWithErr
+    , shellStrictWithErr
 
     -- * Permissions
     , Permissions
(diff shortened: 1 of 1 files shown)


devonhollowood:
I added a couple new functions which seem to be in line with the consensus in #139, as far as I can tell. I did it in terms of a new function, `systemStrictWithErr`, which is very similar to `systemStrict`--it just captures `stderr` as well. These two functions have a lot of code in common, so if you want me to factor this common code into a third function, let me know.

Gabriel439:
Great job!  Thank you so much for contributing this :)

Issue #162, opened 2016-07-18
grep (and sed?) operate on whole Text contents, not individual lines.


unknownloner:
Ran into this while trying to process some text I had generated outside of Shell.

Basically, I wrote `grep pattern (pure x)` where x is a multi-line Text value. Rather than getting only the lines containing `pattern`, I got the entire contents of `x` (or none of the contents if the pattern didn't match of course). This also broke things like `begins` and `ends`, since they were searching from the start/end of the Text, rather than each line.

To work around this I did `grep pattern . cat . map pure . lines $ x` which seems a bit clunky to say the least.

I know that the typical Shell Text producers like readTextFile generate the Text line by line, but I think grep/sed should probably also handle multi-line for cases like this.

Quick edit:
If the rest of Turtle also operates under the assumption that Text values only contain a single line, I realize this could present a bigger problem than just changing grep / sed. Maybe there should be another way to embed Text which handles splitting up the lines for the user, and warnings against using pure/return? I don't really have an answer there.

Gabriel439:
Yeah, the library operates under the assumption that `Text` values only contain a single line.

A shorter way to convert to a `Shell` is:

```haskell
select . lines :: Text -> Shell Text
```

... or when combined with `grep`:

```haskell
grep pattern (select (lines x))
```

Gabriel439:
I'll close this question since I believe it was answered, but you can reopen this if you still have any issues

int-index:
@Gabriel439 Do you think it would be useful to have a wrapper `newtype Line = Line Text` that maintains the invariant?

Gabriel439:
@int-index Yeah.  Let's open a pull request so that people can comment on this change, but I think it would help a lot to avoid this common source of confusion

PR #179, opened 2016-09-17
One-line text newtype

int-index:
A sketch of a newtype wrapper for text values that do not contain newlines. I'm not sure where the assumption about newlines is made in the library and where it isn't, so I didn't use `Line` anywhere except `grep` as per #162.

### DIFF
--- a/src/Turtle/Line.hs
+++ b/src/Turtle/Line.hs
@@ -1,0 +1,97 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DeriveDataTypeable         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Turtle.Line
+  ( Line
+  , lineToText
+  , textToLines
+  , linesToText
+  , textToLine
+  , unsafeTextToLine
+  , NewlineForbidden(..)
+  ) where
+
+import Data.Text (Text)
+import qualified Data.Text as Text
+#if __GLASGOW_HASKELL__ >= 708
+import Data.Coerce
+#endif
+import Data.String
+#if __GLASGOW_HASKELL__ >= 710
+#else
+import Data.Monoid
+#endif
+import Data.Maybe
+import Data.Typeable
+import Control.Exception
+
+-- | The `NewlineForbidden` exception is thrown when you construct a `Line`
+-- using an overloaded string literal or by calling `fromString` explicitly
+-- and the supplied string contains newlines. This is a programming error to
+-- do so: if you aren't sure that the input string is newline-free, do not
+-- rely on the @`IsString` `Line`@ instance.
+--
+-- When debugging, it might be useful to look for implicit invocations of
+-- `fromString` for `Line`:
+--
+-- >>> sh (do { line <- "Hello\nWorld"; echo line })
+-- *** Exception: NewlineForbidden
+--
+-- In the above example, `echo` expects its argument to be a `Line`, thus
+-- @line :: `Line`@. Since we bind @line@ in `Shell`, the string literal
+-- @\"Hello\\nWorld\"@ has type @`Shell` `Line`@. The
+-- @`IsString` (`Shell` `Line`)@ instance delegates the construction of a
+-- `Line` to the @`IsString` `Line`@ instance, where the exception is thrown.
+--
+-- To fix the problem, use `textToLines`:
+--
+-- >>> sh (do { line <- select (textToLines "Hello\nWorld"); echo line })
+-- Hello
+-- World
+data NewlineForbidden = NewlineForbidden
+  deriving (Show, Typeable)
+
+instance Exception NewlineForbidden
+
+-- | A line of text (does not contain newlines).
+newtype Line = Line Text
+  deriving (Eq, Ord, Show, Monoid)
+
+instance IsString Line where
+  fromString = fromMaybe (throw NewlineForbidden) . textToLine . fromString
+
+-- | Convert a line to a text value.
+lineToText :: Line -> Text
+lineToText (Line t) = t
+
+-- | Split text into lines. The inverse of `linesToText`.
+textToLines :: Text -> [Line]
+textToLines =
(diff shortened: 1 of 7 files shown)


int-index:
So, where else should be `Line` used?

src/Turtle/Line.hs
```diff
@@ -0,0 +1,71 @@
...
+#if __GLASGOW_HASKELL__ >= 708
+      coerce
+#else
+      map lineToText
+#endif
+
+-- | Precondition (checked): the argument does not contain newlines.
+-- Throws: `NewlineForbidden`.
+-- You probably want `textToLines` instead.
+textToLine :: Text -> Line
```
Gabriel439: I feel like this function should return `Maybe` and only the `fromString` instance should `throw`

Gabriel439:
The functions that should probably use `Line` are:

* `echo`
* `err`
* `inproc`
* `inshell`
* `inprocWithErr`
* `inshellWithErr`
* `stdin`
* `input`
* `inhandle`
* `stdout`
* `output`
* `outhandle`
* `append`
* `stderr`
* `strict`
* `sed`
* `yes`
* `countChars`
* `countLines`
* `countWords`
* `proc`
* `shell`
* `system`
* `procs`
* `shells`
* `procStrict`
* `shellStrict`

I will try to build [downstream packages](http://packdeps.haskellers.com/reverse/turtle) against this pull request to see how disruptive the change is to their code

src/Turtle/Line.hs
```diff
@@ -0,0 +1,71 @@
...
+    mapTextToLine :: [Text] -> [Line]
+    mapTextToLine =
+#if __GLASGOW_HASKELL__ >= 708
+      coerce
+#else
+      map unsafeTextToLine
+#endif
+
+linesToText :: [Line] -> Text
+linesToText = Text.unlines . mapLineToText
```
Gabriel439: You can `coerce` the `Text.unlines` function directly, like this:

```haskell
linesToText :: [Line] -> Text
linesToText =
#if __GLASGOW_HASKELL__ >= 708
    coerce Text.unlines
#else
    map lineToText . Text.unlines
#endif

**Comment by int-index** [comment]
Current progress:

* [ ] `echo`
* [ ] `err`
* [ ] `inproc`
* [ ] `inshell`
* [ ] `inprocWithErr`
* [ ] `inshellWithErr`
* [ ] `stdin`
* [ ] `input`
* [ ] `inhandle`
* [ ] `stdout`
* [ ] `output`
* [ ] `outhandle`
* [ ] `append`
* [ ] `stderr`
* [ ] `strict`
* [ ] `sed`
* [ ] `yes`
* [ ] `countChars`
* [ ] `countLines`
* [ ] `countWords`
* [ ] `proc`
* [ ] `shell`
* [ ] `system`
* [ ] `procs`
* [ ] `shells`
* [ ] `procStrict`
* [ ] `shellStrict`

**Comment by int-index** [comment]
Should `sed` take a `Pattern Line` or `Pattern Text`?

File: src/Turtle/Line.hs (hunk 1 above)
Comment (int-index) [inline src/Turtle/Line.hs]: Done.

File: src/Turtle/Line.hs (hunk 2 above)
Comment (int-index) [inline src/Turtle/Line.hs]: Done.

**Comment by Gabriel439** [comment]
@int-index `sed` should definitely take a `Pattern Text` as an argument.  The result of the `Pattern` is used to replace the substring it matches, which is not necessarily the entire line

**Comment by int-index** [comment]
Can it contain newlines? If not, it's a `Line`.

**Comment by int-index** [comment]
The problem is that I don't know what to do if the result of matching against the pattern is a `Text` value with newlines. Should I yield each line separately?

**Comment by Gabriel439** [comment]
Yeah, I think it would make sense to split the resulting of the substitution into lines and yield each line separately

File: src/Turtle/Prelude.hs [hunk 3]
```diff
@@ -1415,14 +1416,15 @@ grep pattern s = do
...
 sed pattern s = do
     when (matchesEmpty pattern) (die message)
     let pattern' = fmap Text.concat
             (many (pattern <|> fmap Text.singleton anyChar))
-    txt    <- s
-    txt':_ <- return (match pattern' txt)
-    return txt'
+    line   <- s
+    txt':_ <- return (match pattern' (lineToText line))
+    line' <- select (textToLines txt')
```
Gabriel439: You can simplify this:

```haskell
line' <- select (textToLines txt')
return line'
```

... to this:

```haskell
select (textToLines txt')
```

int-index: In my defense - I was sleepy when I wrote this :smile: 
Fixed.

int-index:
@Gabriel439 Do you think I can merge this or are there other problems with the PR?

src/Turtle/Line.hs
```diff
@@ -0,0 +1,65 @@
...
+import Data.String
+#if __GLASGOW_HASKELL__ >= 710
+#else
+import Data.Monoid
+#endif
+import Data.Maybe
+import Data.Typeable
+import Control.Exception
+
+data NewlineForbidden = NewlineForbidden
```
Gabriel439: Could you add a comment to this newtype explaining what it means to receive this exception?

Gabriel439:
Before merging this I would like to go through the tutorial and see how it affects all the examples to see if there are any rough edges with this new API

src/Turtle/Line.hs
int-index: I added a comment that explains what may cause the exception and how to debug it.

int-index:
@Gabriel439 I checked all examples from the tutorial and updated the ones broken by the change. Mostly it's just type signature changes, but there's a problem that `echo` can no longer be used with `format` (since there's no guarantee that the result of `format` is a single line). It's easy to side-step by using `printf` instead.

Gabriel439:
@int-index Awesome!  One last request: could you bump the version to `1.3.0` and then feel free to merge :)

PR #180, opened 2016-09-18
Add onFiles utility

devonhollowood:
Implement #149. This version ignores files which do not decode as `Text` properly. Instead, we could use the approximation of the `FilePath` given by `toText`'s `Left`. Let me know if you want me to switch the behavior to that, or if you want the behavior to be something else entirely.


Gabriel439:
Looks great!  Thanks :)

PR #183, opened 2016-09-27
Show help on parse error/empty command

mitchellwrosen:


Gabriel439:
Nice idea!  Thanks :)

PR #185, opened 2016-09-29
Add subcommandGroup

mitchellwrosen:
The code came out a bit awkward because the underlying API is awkward. The entire command group is given a metavar (rather than each subcommand), so I just manually emulate the metavar generated by `<|>` (this is also the reason I took the entire command group as a monolithic list, rather than use `<>` to build one up)

### DIFF
--- a/turtle.cabal
+++ b/turtle.cabal
@@ -64,7 +64,7 @@ Library
         text                               < 1.3 ,
         time                               < 1.7 ,
         transformers         >= 0.2.0.0 && < 0.6 ,
-        optparse-applicative >= 0.11    && < 0.14,
+        optparse-applicative >= 0.13    && < 0.14,
         optional-args        >= 1.0     && < 2.0 ,
         unix-compat          >= 0.4     && < 0.5
     if os(windows)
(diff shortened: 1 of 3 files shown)


Gabriel439:
I like this!  I will do some work to fix the Travis build errors so that we can merge this

turtle.cabal
```diff
@@ -64,7 +64,7 @@ Library
         text                               < 1.3 ,
         time                               < 1.7 ,
         transformers         >= 0.2.0.0 && < 0.6 ,
-        optparse-applicative >= 0.11    && < 0.14,
```
Gabriel439: This is the source of the build failures in CI.  The root of the problem is that `criterion-1.1.1.0` has an `optparse-applicative < 0.13.0.0` upper bound so the benchmark suite can't build.  There is an issue open on Stackage to fix this:

https://github.com/fpco/stackage/issues/1784

We need to wait on that to be fixed before we can merge this

domenkozar: It's done: https://github.com/fpco/stackage/issues/1784#issuecomment-258897198

Gabriel439:
Thanks!  The upstream problem was fixed so this is clear to merge now.  I'm sorry for the long delay

PR #188, opened 2016-10-22
Added functions to help out with headers in streaming.

GregorySchwartz:
I have huge needs for streaming functions that take the first line across the stream for data processing. For instance, filtering based on a column name of a csv file, adding columns, columns abound, so I believe headerMap and headerRMap can fill this need. For instance (in an overly wordy way as it's a general function, not necessarily for text or the like),

> view (headerRMap (\h l -> mappend (h !! 3) (l !! 2)) . fmap (T.splitOn "," . lineToText) $ Turtle.input "test.csv")  
> Just "testa"
> Just "test4"
> Just "testalfred"

It would be fun to implement, in addition, a version meant solely for csv files, but that would require more dependencies (to make my life easier) which you would probably not want. For now, these functions would be very useful to have in the Prelude.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -189,6 +189,8 @@ module Turtle.Prelude (
 
     -- * Text
     , cut
+    , headerMap
+    , headerRMap
 
     -- * Subprocess management
     , proc
@@ -1622,6 +1624,25 @@ cache file s = do
 cut :: Pattern a -> Text -> [Text]
 cut pattern txt = head (match (selfless chars `sepBy` pattern) txt)
 -- This `head` should be safe ... in theory
+
+-- | Map a function that uses the header over the stream.
+headerMap :: (a -> a -> a) -> Shell a -> Shell (Maybe a)
+headerMap f s = do
+    header <- fold s Control.Foldl.head
+    case header of
+        Nothing  -> return Nothing
+        (Just h) -> cat [ return . Just $ h
+                        , fmap (Just . f h) s
+                        ]
+
+-- | Map a function that uses the header over the stream, removing the header,
+-- allowing a more flexibility.
+headerRMap :: (a -> a -> b) -> Shell a -> Shell (Maybe b)
+headerRMap f s = do
+    header <- fold s Control.Foldl.head
+    case header of
+        Nothing -> return Nothing
+        (Just h) -> fmap (Just . f h) $ s
 
 -- | Get the current time
 date :: MonadIO io => io UTCTime


Gabriel439:
Ideally you want to go over the stream once in a single pass when possible.  Here's an example of how you can efficiently do that and get rid of the `Maybe` from the result type:

```haskell
import Control.Foldl (FoldM(..))
import Turtle (Shell(..))

withHeader :: (a -> a -> b) -> Shell a -> Shell b
withHeader f (Shell k) = Shell k'
  where
    -- In practice this would use a strict pair and strict `Maybe` to avoid a space leak
    k' (FoldM step begin done) = k (FoldM step' begin' done')
      where
        step' (x, Nothing) a = do
            return (x, Just (f a))
        step' (x, Just g ) a = do
            x' <- step x (g a)
            return (x', Just g)

        begin' = do
            x <- begin
            return (x, Nothing)

        done' (x, _) = done x
```

Example usage:

```haskell
>>> view (withHeader (,) (select [0..10]))
(0,1)
(0,2)
(0,3)
(0,4)
(0,5)
(0,6)
(0,7)
(0,8)
(0,9)
(0,10)
```

Would that solve your use case?

GregorySchwartz:
Ah, yes, my unfamiliarity with Shell is showing...

That result looks good, but there should also be a version, in addition, that retains the header with the restriction that the header and body are of the same type, so you can, for instance, have a csv file where you want to add 1 to a column (by header name). In that case, you would have the header be a string and the body be a string as well, so you can retain the header after the map.

Thinking about it, having a filter version would be wonderful as well, so you could filter by the header (like getting values greater than 3 in a certain column by name). It would be exactly the same but skipping based on a predicate rather than applying a function.

Gabriel439:
So my understanding is that if I slightly modify this to:

```haskell
header :: Shell a -> Shell (a, a)
```

... where the left element of the tuple is the header (always the same taken from the first line) and the second element of the tuple is a non-header line, then this would solve your problem.

For example, if you wanted to do something to each line based on the header, you would write:

```haskell
-- `transform` is a function that take the header and the current line and returns an updated line
transform :: Line -> Line -> Line
transform header currentLine = ...

example = do
    (header, currentLine) <- header (input "someFile")
    return (transform header currentLine)
```

Similarly, if you wanted to filter lines based on the header, you would write:

```haskell
-- `keep` is a function that takes the header and the current line and decides whether or not
-- to keep the line.  `True` means to keep the line
keep :: Line -> Line -> Bool
keep header currentLine = ...

example = do
    (header, currentLine) <- header (input "someFile")
    guard (keep header currentLine)
    return currentLine
```

Would that solve your use cases?

GregorySchwartz:
But then there would no longer be any streaming and easy composition with other shell functions, right?

Gabriel439:
Yes, this would still stream and compose with other shell functions.  Here's the implementation I had in mind:

```haskell
import Control.Foldl (FoldM(..))
import Turtle (Shell(..))

data Pair a b = Pair !a !b

header :: Shell a -> Shell (a, a)
header (Shell k) = Shell k'
  where
    -- In practice this would use a strict pair and strict `Maybe` to avoid a space leak
    k' (FoldM step begin done) = k (FoldM step' begin' done')
      where
        step' (Pair x  Nothing ) a = do
            return (Pair x (Just ((,) a)))
        step' (Pair x (Just g )) a = do
            x' <- step x (g a)
            return (Pair x' (Just g))

        begin' = do
            x <- begin
            return (Pair x Nothing)

        done' (Pair x _) = done x
```

... and you can prove to yourself that this streams in constant space by just running:

```
>>> view (header (select [1..]))
(1,2)
(1,3)
(1,4)
...
```

GregorySchwartz:
I mean, what would the composable and streamable version be if you wanted to map AND keep the line? That is, something like
```
>>> view (headerKeep (+) (select [1..]))
1
3
4
5
...
```

Gabriel439:
Oh, I see what you mean now.  Perhaps the type signature could be:

```haskell
data WithHeader a
    = Header a
    -- ^ The first line with the header
    | Row a a
    -- ^ Every other line: first element is header, second element is original row

header :: Shell a -> Shell (WithHeader a)
```

... then your example would be:

```haskell
example = do
    x <- header (select [1..])
    case x of
        Header n -> return n
        Row x y -> return (x + y)
```

Would that work for you?

GregorySchwartz:
Yes, much better, as that last line could be `Row x y -> return (if x == "howdy" || y == "cheese" then 3 else 10)` would work as well, or it could be whatever type you want, and then having the option of the Header or just ignoring it. Perfect!

Issue #102, opened 2015-09-01
lnghc: fd:15: hClose: resource vanished (Broken pipe)


epsilonhalbe:
Hey I think I found a bug in the Turtle-Library.

I wrote the following script in order to easily switch between the ghc-versions I use (provided by [hvr's-ppa repository](https://launchpad.net/~hvr/+archive/ubuntu/ghc)).

```haskell
{-# LANGUAGE OverloadedStrings #-}

module Main where

import Turtle
import Prelude hiding (FilePath)
import Data.Either (rights)
import qualified Control.Foldl as F
import qualified Data.Text as T
import Data.Text()

main :: IO ()
main = sh $ do
          args <- arguments
          available_ghcs <- map filename <$> (fold (ls "/opt/ghc/") F.list)
          Just home <- need "HOME"
          case args of
            (a:_) -> if (fromText a) `elem` available_ghcs
                        then do _ <- unlink  $ home <> "/bin/ghc"
                                _ <- symlink ("/opt/ghc/" <> a <> "/bin") $ home <> "/bin/ghc"
                                echo $ "...successfully changed GHC ["<>a<>"]"
                        else do echo $ "'"<> a <>"' is no valid GHC version"
                                list available_ghcs
            _ -> do echo "USAGE: lnghc ${GHC VERSION}"
                    list available_ghcs
    where unlink trgt = proc "unlink" [trgt] ""
          symlink src trgt = proc "ln" ["-s",src,trgt] ""


list :: [FilePath] -> Shell ()
list versions = do current_ghc <- lastMay . T.words <$> inproc "ghc" ["--version"] ""
                   echo "available GHC versions: "
                   mapM_ (echo . highlight current_ghc) $ rights $ map toText versions



highlight :: Maybe Text -> Text -> Text
highlight hi str = if hi == Just str
                     then "[" <> str <> "]"
                     else " " <> str <> " "

lastMay :: [a] -> Maybe a
lastMay [] = Nothing
lastMay xx = Just $ last xx
```

the corresponding cabal file contains

```
...
  build-depends:       base >= 4.7 && < 5
                     , turtle
                     , foldl
                     , text
                     , system-filepath
...
```
and I've used stack with the current resolver (lts-3.0) and ghc version 7.10.2 to compile it.

When running the script multiple times I get the error above - I think the culprit may be the `unlink`/`link` usages of `proc`.

If I can help fixing this somehow, I will be glad to do so.

KR Martin

[...]

PR #191, opened 2016-11-16
Fix broken pipe issue with proc/shell functions

maoe:
Relevant: #102

The first commit [demonstrates](https://travis-ci.org/maoe/Haskell-Turtle-Library/jobs/176547633#L2338) the issue.

Here is what's happening:
1. `system` creates a pipe to feed input to the command it's executing
2. the command exits immediately, so the pipe is closed
3. `feedIn` tries to `write(2)` to the closed pipe, which ends up with broken pipe

Note that `hClose` also can cause broken pipe as it flushes the buffer before closing the handle.

The issue can be worked around by ignoring `SIGPIPE` as the `process` package does in `readCreateProcess` etc.

PS: I'm not sure why `turtle` needs to call `close` twice in `system` and the like. Isn't it sufficient to call it once in `feedIn`?

### DIFF
--- a/src/Turtle/Bytes.hs
+++ b/src/Turtle/Bytes.hs
@@ -44,6 +44,7 @@ import Filesystem.Path (FilePath)
 import Prelude hiding (FilePath)
 import System.Exit (ExitCode(..))
 import System.IO (Handle)
+import Turtle.Internal (ignoreSIGPIPE)
 import Turtle.Prelude (ProcFailed(..), ShellFailed(..))
 import Turtle.Shell (Shell(..), fold, sh)
 
@@ -334,7 +335,8 @@ system p s = liftIO (do
     mvar <- MVar.newMVar False
     let close handle = do
             MVar.modifyMVar_ mvar (\finalized -> do
-                Control.Monad.unless finalized (System.IO.hClose handle)
+                Control.Monad.unless finalized
+                    (ignoreSIGPIPE (System.IO.hClose handle))
                 return True )
     let close' (Just hIn, ph) = do
             close hIn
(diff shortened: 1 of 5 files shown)


maoe:
One of the transitive dependencies (`js-jquery`) failed to build on Tranvis with GHC 7.6.3.

Gabriel439:
The `js-jquery` issue is now fixed in `master` so if you merge `master` the build failure will disappear

I believe the second `hClose` in `close'` is still technically necessary under very rare circumstances.  There is a brief moment before the `mask_` begins that an asynchronous exception could occur

Could you make the same modification to the `Turtle.Bytes` module?

maoe:
Do you have any preference as to where to put `ignoreSIGPIPE`? I guess you'd like to create a separate internal module for this so both of `Tutle.Prelude` and `Turtle.Bytes` can use the same function.

Re: `outhandle` in #197, I'm not sure if we really want to upstream `ignoreSIGPIPE` into `outhandle`. `outhandle` doesn't always necessarily need to ignore SIGPIPE, right?

Regarding the `hClose`, sorry I misread the code. Yes, we definitely need to call it in the bracket.

int-index:
Will it work if we add `ignoreSIGPIPE` this way? There's no need to use it inside `outhandle`.

```
feedIn restore = restore (ignoreSIGPIPE (outhandle hIn s)) `finally` close hIn
--                        ^^^^^^^^^^^^^ here
```

maoe:
Sure, that would work.

maoe:
Rebased off of current master.

I also applied the outhandle changes and the broken pipe fix to `Turtle.Bytes`.

Gabriel439:
Looks great!  Thank you for fixing this :)

PR #192, opened 2016-11-17
More flexible command line flags

maoe:
This PR changes the way to define command line flags to get more flexibility. Specifically the option parsing library now:

* requires only one flag name which can be a short name or a long name
* is able to define multiple long/short names for a flag e.g. `switch ("dry" <> "dry-run" <> shortFlag 'n')` accepts `--dry`, `--dry-run`, and `-n`
* requires a separate metavariable name. This is necessary because there can be no long flag name which can be used as the metavariable. This is sometimes useful even with a long flag for clarity of the help message.

## Motivation

There are three reasons why I want this change.

1. I often would like to omit short flags for rarely-used options. This is currently impossible because all `optXXX` parsers always take an `ArgName` and a `ShortName`.
2. I sometimes would like to define a synonym for a flag so I can change a flag name without breaking existing systems.
3. I'd like to specify a custom metavariable if the flag takes some complex format.

## Implementation notes

`FlagName` is an abstract data type accompanied by smart constructors to keep an invariant: every option should at least have either a long name or a short name. It has an instance of `Semigroup` so we can add more names without introducing `mempty` which breaks the invariant.

## Migration notes

This is a breaking change. All existing code that use `optXXX` break. Fortunately, it's straightforward to fix the breakage:

* `ArgName` is a long option. If it's a string literal, it works as it is. Otherwise you can use `longFlag`.
* `shortFlag` is necessary to turn a `ShortName` to a `FlagName`. You can compose it with other flag names using `(Data.Semigroup.<>)`, which is re-exported from `Turtle`.
* `ArgName` now means the metavariable for the flag.

Look at the documentation changes in this PR for example.

Another type of breakage is the reexport of `Semigroup`'s `(<>)` from `Turtle`. It fails to compile if a module imports `(<>)` from `Data.Monoid`. I suspect it would become less problematic as `Data.Semigroup` gets popular.

### DIFF
--- a/src/Turtle.hs
+++ b/src/Turtle.hs
@@ -76,6 +76,7 @@ module Turtle (
     , module Control.Monad
     , module Control.Monad.IO.Class
     , module Data.Monoid
+    , module Data.Semigroup
     , module Control.Monad.Managed
     , module Filesystem.Path.CurrentOS
     , Fold(..)
@@ -117,7 +118,8 @@ import Control.Monad
     , unless
     )
 import Control.Monad.IO.Class (MonadIO(..))
-import Data.Monoid (Monoid(..), (<>))
+import Data.Monoid (Monoid(..))
+import Data.Semigroup (Semigroup(..))
 import Data.String (IsString(..))
 import Filesystem.Path.CurrentOS
     ( FilePath
--- a/src/Turtle/Line.hs
+++ b/src/Turtle/Line.hs
@@ -24,6 +24,7 @@ import Data.String
 import Data.Monoid
 #endif
 import Data.Maybe
+import Data.Semigroup
 import Data.Typeable
 import Control.Exception
 
@@ -57,7 +58,7 @@ instance Exception NewlineForbidden
 
 -- | A line of text (does not contain newlines).
 newtype Line = Line Text
-  deriving (Eq, Ord, Show, Monoid)
+  deriving (Eq, Ord, Show, Monoid, Semigroup)
 
 instance IsString Line where
   fromString = fromMaybe (throw NewlineForbidden) . textToLine . fromString
--- a/src/Turtle/Options.hs
+++ b/src/Turtle/Options.hs
@@ -10,8 +10,8 @@
 -- > import Turtle
 -- >
 -- > parser :: Parser (Text, Int)
--- > parser = (,) <$> optText "name" 'n' "Your first name"
--- >              <*> optInt  "age"  'a' "Your current age"
+-- > parser = (,) <$> optText (longFlag "name" <> shortFlag 'n') "name" "Your first name"
+-- >              <*> optInt  (longFlag "age")                   "age"  "Your current age"
 -- >
 -- > main = do
 -- >     (name, age) <- options "Greeting script" parser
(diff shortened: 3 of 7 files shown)


Gabriel439:
The main issue here is that as a rule I try to ensure that types with an `IsString` and `Semigroup` instance should satisfy the following law:

```haskell
fromString (x <> y) = fromString x <> fromString y
``` 

If I were to make a breaking change I would prefer to make both the `LongName` and `ShortName` into `Optional` arguments.  For the `LongName` this would have low impact to the user since `Optional` has an `IsString` instance, but users would have to wrap the `ShortName` argument in `pure` or use `empty` to omit it.

As a side note: I wish Haskell had a way to overload character literals so that we could do something like:

```haskell
instance IsChar a => IsChar (Optional a) where
    fromChar c = pure (fromChar c)
```

maoe:
I'm fine to drop the `IsString` instance for `FlagName` if that's better.

I think there are some design choices here.

My implementation ensures there is at least one name for an option. This is hard to guarantee if we just make `LongName` and `ShortName` into `Optional` arguments in `optXXX` and `switch` because it's possible that both names are empty. In that case, we could make the `ArgName` the default option name. It forces users to make at least one long name for an option, which is, I feel, an unnecessary constraint.

Alternatively we could use `NonEmpty FlagName`  where `data FlagName = ShortName Char | LongName Text`. This satisfies the same requirement (where every option has at least one name, either long one or short one) but with a different syntax. I chose `Semigroup` because `Data.List.NonEmpty` is a bit tedious to use.

Lastly, I agree that `OverloadedCharacters` would be nice.

Gabriel439:
The other issue with `Semigroup` is that it permits things like this:

```haskell
longName "foo" <> longName "bar"
```

How about just making `ShortName` an `Optional` argument (but not `LongName`)?

I can also submit a proposal to see if people are open to an `OverloadedCharacters` extension

maoe:
I've just rebased my branch off of current master. No code changes.

> longName "foo" <> longName "bar"

Hmm is that bad? I [intentionally made `FlagName` accept multiple names](https://github.com/Gabriel439/Haskell-Turtle-Library/pull/192/commits/af1ca2a2982685fff818f3529e9fc4f053ab3418) because it's useful to define flag synonyms as described in the motivation section.

> How about just making ShortName an Optional argument (but not LongName)?

This means we always need a `LongName`.

What I'm trying here is to make `turtle` cover common patterns when writing command line parsers:

* A flag can have multiple names (synonyms). e.g. `tar` accepts `--extract` *and* `--get` for the same thing.
* A flag can only have either a short name or a long name. e.g. `ls -1` doesn't have corresponding long names, `ls --show-control-chars` doesn't have corresponding short names.

The approach of `LongName` and `Optional ShortName` doesn't work well for these cases.

Gabriel439:
I would still prefer to keep this module the way it is.  The goal of this module is to be opinionated and it doesn't have to cover all possible use cases.  Also, the user can still use `optparse-applicative` directly for any use case not covered by this module

maoe:
> The goal of this module is to be opinionated and it doesn't have to cover all possible use cases. Also, the user can still use optparse-applicative directly for any use case not covered by this module

Fair enough. I'll use `optparse-applicative` directly where needed.

Thanks for the reviews.

Gabriel439:
You're welcome!

PR #194, opened 2016-11-18
Return an empty Line if textToLine is applied to an empty Text

maoe:
Currently the `IsString` instance for `Line` doesn't accept an empty string as an input. This is a bit weird as, I think, an empty string is completely valid data to be used as a `Line`. This is in practice a source of partiality:

```
*Turtle> :set -XOverloadedStrings 
*Turtle> echo ""
*** Exception: NewlineForbidden
```

This PR fixes `textToLine` to return an empty `Line` if it's applied to an empty `Text`.


maoe:
Updated the example above and fixed a bug in the implementation.

Gabriel439:
Thank you for fixing this!  I agree that `""` should be a valid `Line`

PR #196, opened 2016-11-19
Add parallel utility with the help of fork+wait

### LINKED ISSUE #21: Feature request: parallel
It would be great, to have the facility to run e.g. external commands in parallel.

domenkozar:
Fixes #21

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -181,6 +181,7 @@ module Turtle.Prelude (
     , limit
     , limitWhile
     , cache
+    , parallel
 
     -- * Folds
     , countChars
@@ -265,7 +266,7 @@ import Control.Exception (Exception, bracket, bracket_, finally, mask_, throwIO)
 import Control.Foldl (Fold, FoldM(..), genericLength, handles, list, premap)
 import qualified Control.Foldl
 import qualified Control.Foldl.Text
-import Control.Monad (guard, liftM, msum, when, unless)
+import Control.Monad (guard, liftM, msum, when, unless, (>=>))
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Managed (MonadManaged(..), managed, managed_, runManaged)
 #ifdef mingw32_HOST_OS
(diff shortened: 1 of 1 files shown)


src/Turtle/Prelude.hs
```diff
@@ -1331,6 +1332,28 @@ fork io = using (managed (withAsync io))
...
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+Sat Nov 19 17:26:52 CET 2016
+
+-}
+parallel :: MonadManaged managed => [IO ()] -> managed ()
```
Gabriel439: You could actually probably simplify the type to:

```haskell
parallel :: [IO ()] -> IO ()
```

... by just wrapping the whole thing in `runManaged` since you don't really benefit from keeping the result inside of `Managed`.  However, then this wouldn't be really `turtle`-specific at all and I'm pretty sure there's a library on Hackage that already provides such a function

Also, I think for this to be more in the spirit of the GNU `parallel` command you'd want to be able to collect the results so they could be processed further, using a type like this:

```haskell
-- | Fork all the `IO` actions and stream their results as they complete, preserving order
parallel :: [IO a] -> Shell a
```

... and then if you want to ignore their results, you would just wrap the whole thing in `sh`.  You could implement it something like this:

```haskell
parallel :: [IO a] -> Shell a
parallel ios = do
    asyncs <- mapM fork ios
    async  <- select asyncs
    wait async
```

So, for example, this code:

```haskell
main = view (parallel (map sleep [1, 2, 3, 1, 2, 1]))
```

... would print:

```haskell
() -- After 1 second
() -- After 2 seconds
() -- After 3 seconds
() -- After 3 seconds
() -- After 3 seconds
() -- After 3 seconds
```

Gabriel439:
Also, if merge `master` it has a fix for the CI build failure

src/Turtle/Prelude.hs
int-index: Or more succinctly, `traverse fork >=> select >=> wait`.

domenkozar: @Gabriel439 so would you prefer `parallel = traverse fork >=> select >=> wait`?

Gabriel439: Yeah

3noch: But can we do better?? 😉 

> If I had more time, I would have written a shorter letter.

\- [Blaise Pascal](http://quoteinvestigator.com/2012/04/28/shorter-letter/)

domenkozar:
@Gabriel439 @int-index pushed

Gabriel439:
Thank you!

Issue #167, opened 2016-08-24
find is so slow


3noch:
I'm trying to replicate the simple logic of `find . -name "*.orig" | xargs rm`, but my _compiled_ (with `-O2`) script takes 20 seconds as opposed to instantaneous. It's possible that `Turtle.find` is doing more work by recursing into hidden directories or something, but the performance is so stark I wonder if we need to at least document something:

``` haskell
#!/usr/bin/env stack
-- stack --install-ghc runghc --package turtle


{-# LANGUAGE OverloadedStrings #-}

import Turtle

main = do
    extension <- options "Remove all files with an extension" $
      optText "extension" 'e' "File extension to remove"

    sh $ do
      doomed <- find (ends $ text $ "." <> extension) "."
      liftIO $ do
        printf ("Removing "%fp%"\n") doomed
        --rm doomed

```

3noch:
I've confirmed that Turtle's `find` is finding far more files than Linux `find`.

```shell
(climate) ~/a/l/f/climate ❯❯❯ find . -name "*" | wc -l
    8173
(climate) ~/a/l/f/climate ❯❯❯ turtlefind | wc -l
  384676
```

mitchellwrosen:
Turtle's `find` does recurse into hidden directories, but so does `find`.

mitchellwrosen:
@3noch, compare find's output to `turtlefind | sed 's/FilePath /' | tr -d '"'` and see what the difference is. I tried your experiment and I'm seeing the same files output by both tools (except that `find` outputs `.` and turtle does not)

3noch:
I wonder if it could by symlinks.

mitchellwrosen:
Ah, yes. `find` will list a directory that is a symlink, but not go into it.

mitchellwrosen:
This is easy to fix - just change `lstree` to `lsif fileIsSymbolicLink`. Problem is, there's currently CPP that partitions windows/posix code for this case. I have a PR up that adds `unix-compat` as a dependency - if Gabriel wants to merge it, then this change will be able to leverage that library's `getSymbolicLinkStatus`.

Gabriel439:
Yes, I'm totally fine with adding `unix-compat` as a dependency, since I believe we will need it for all symlink-related stuff

Gabriel439:
Note that there are two possible reasons that `find` might be slow:

* differences in which files are traversed, as already discussed
[...]

PR #202, opened 2016-12-18
Don't follow symlinks in 'find'

mitchellwrosen:
Addresses (part of) #167. Makes `find` act like `find -P`, which is the default.


Gabriel439:
Thanks!

PR #203, opened 2016-12-22
Add fpToLine for converting from FilePath to Line directly

PierreR:
I had to write some alike : 
```
unless found $ do
  liftIO $ writeTextFile configfile "..."
  echo $ unsafeTextToLine $ format (fp%" created") configfile 
```
The last line looks like a lot of boilerplate when in bash it is simply `echo "${configfile} created"`. 

That has been the incentive to write 'fpToLine`.

Does it look like a valuable addition ?

### DIFF
--- a/src/Turtle/Line.hs
+++ b/src/Turtle/Line.hs
@@ -10,6 +10,7 @@ module Turtle.Line
   , linesToText
   , textToLine
   , unsafeTextToLine
+  , fpToLine
   , NewlineForbidden(..)
   ) where
 
@@ -26,6 +27,7 @@ import Data.Monoid
 import Data.Maybe
 import Data.Typeable
 import Control.Exception
+import Filesystem.Path.CurrentOS (FilePath, toText)
 
 -- | The `NewlineForbidden` exception is thrown when you construct a `Line`
 -- using an overloaded string literal or by calling `fromString` explicitly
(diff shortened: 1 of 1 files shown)


src/Turtle/Line.hs
```diff
@@ -10,6 +10,7 @@ module Turtle.Line
   , linesToText
   , textToLine
   , unsafeTextToLine
+  , fpToLine
```
Gabriel439: How about calling this `pathToLine`?

int-index:
> echo $ unsafeTextToLine $ format (fp%" created") configfile 

I believe this is equivalent to `printf (fp%" created") configfile`.

int-index:
Also, according to http://superuser.com/questions/129519/which-file-systems-support-newlines-in-file-names, the function you're proposing is not safe — filenames can contain newlines. I've just tested this on my machine and it's true:

```
Prelude Turtle> sh $ touch "hello\nworld"
Prelude Turtle> view $ ls "."
FilePath "./hello\nworld"
```

PierreR:
> I believe this is equivalent to printf (fp%" created") configfile.

@int-index It does indeed. I didn't know about `Turtle.Format.printf`. Thanks for the tip !

Isn't it a bit confusing to have so many functions to print to `stdout` ;-)

I am kind of flabbergast that know that filenames can have newlines ...

domenkozar:
This change is really painful. I'm now just using `Data.Text.IO.putStrLn`.

mitchellwrosen:
@domenkozar Which change? This PR was closed by doc changes.

PierreR:
As an extra note `printf (fp%" created.\n") configfile` is only equivalent with an extra `\n` .
Otherwise buffering will interact badly with the rest of `turtle` and the line will be emit at the very end (which is probably not want the user wants).

PR #221, opened 2017-03-09
Correctly unmask exceptions when running subprocesses

Gabriel439:
Before this change the following code hangs:

    runManaged (do
        _ <- fork (shells "while true; do sleep 1; done" empty)
        return () )

... because of the following sequence of events:

*   `fork` forks the subprocess using `async` and registers a `cancel` function
    to be called when the `runManaged` block completes
*   `runManaged` completes immediately after that, so `cancel` gets called on
    the thread running the subprocess
*   `cancel` throws a `ThreadKilled` exception to the thread running the
    subprocess
*   `shells` internally uses `system`, which interally uses the following idiom:

    ```haskell
    let feedIn :: (forall a. IO a -> IO a) -> IO ()
        feedIn restore =
            restore (ignoreSIGPIPE (outhandle hIn s)) `finally` close hIn
    mask_ (withAsyncWithUnmask feedIn (\a -> Process.waitForProcess ph <* halt a) )
    ```

    ... which means that `outhandle` runs with exceptions unmasked but the
    `waitForProcess` does not!
*   The `outhandle` function running in the forked `feedIn` thread gets correctly
    terminated by the exception because it runs unmasked inside of the `restore`
*   However, the main thread running `waitForProcess` has exceptions still masked
*   Therefore, the `waitForProcess` does not complete, because it only can
    terminate under two conditions.  Either:
        *   `withAsyncWithUnmask` forwards exceptions from the forked thread to the
            main thread running `waitForProcess` to terminate the main thread
            * However, this doesn't work because the main `waitForProcess` masks
              any exceptions forwarded from the forked thread
        *   `waitForProcess` completes successfully, which only happens if the
            forked thread calls `terminateProcess`
            * However, the forked thread doesn't call `terminateProcess` if it
              fails with an exception

The solution is to correctly unmask the `waitForProcess` so that it correctly
receives any exceptions propagated from the forked thread.  I did this by using
`mask` to provide the `restore` function instead of `withAsyncWithUnmask`.

Another alternative solution would be to have the forked thread call
`terminateProcess` if any exceptions were thrown (which might still be worth
doing in a separate change), but I felt that signaling shutdown by correctly
propagating exceptions is a more reliable mechanism than signaling via the process
handle.

Several other functions use the same idiom, so I updated them to also unmask the
`waitForProcess` call.

I added a regression test for this behavior to ensure that forked subprocesses
terminate immediately.  This test fails before this change and succeeds after.

### DIFF
--- a/src/Turtle/Bytes.hs
+++ b/src/Turtle/Bytes.hs
@@ -349,7 +349,9 @@ system p s = liftIO (do
                 feedIn restore =
                     restore (ignoreSIGPIPE (outhandle hIn s))
                     `Exception.finally` close hIn
-            Exception.mask_ (Async.withAsyncWithUnmask feedIn (\a -> Process.waitForProcess ph <* halt a) )
+            Exception.mask (\restore ->
+                Async.withAsync (feedIn restore) (\a ->
+                    restore (Process.waitForProcess ph) <* halt a ) )
         handle (Nothing , ph) = do
             Process.waitForProcess ph
 
@@ -390,7 +392,9 @@ systemStrict p s = liftIO (do
                 `Exception.finally` close hIn
 
         Async.concurrently
-            (Exception.mask_ (Async.withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* halt a)))
+            (Exception.mask (\restore ->
+                Async.withAsync (feedIn restore) (\a ->
+                    restore (Process.waitForProcess ph) <* halt a ) ))
             (Data.ByteString.hGetContents hOut) ) )
 
 systemStrictWithErr
@@ -428,7 +432,9 @@ systemStrictWithErr p s = liftIO (do
                 `Exception.finally` close hIn
 
         runConcurrently $ (,,)
-            <$> Concurrently (Exception.mask_ (Async.withAsyncWithUnmask feedIn (\a -> liftIO (Process.waitForProcess ph) <* halt a)))
+            <$> Concurrently (Exception.mask (\restore ->
+                    Async.withAsync (feedIn restore) (\a ->
+                        restore (Process.waitForProcess ph) <* halt a ) ))
             <*> Concurrently (Data.ByteString.hGetContents hOut)
             <*> Concurrently (Data.ByteString.hGetContents hErr) ) )
 
@@ -499,7 +505,10 @@ stream p s = do
                 liftIO (Data.ByteString.hPut hIn bytes) ) )
             `Exception.finally` close hIn
 
-    a <- using (Managed.managed (Exception.mask_ . Async.withAsyncWithUnmask feedIn))
+    a <- using
+        (Managed.managed (\k ->
+            Exception.mask (\restore ->
+                Async.withAsync (feedIn restore) k ) ))
     inhandle hOut <|> (liftIO (Process.waitForProcess ph *> halt a) *> empty)
 
 streamWithErr
@@ -563,9 +572,18 @@ streamWithErr p s = do
             x1 <- loop x0 (0 :: Int)
             done x1 )
 
-    a <- using (Managed.managed (Exception.mask_ . Async.withAsyncWithUnmask feedIn    ))
-    b <- using (Managed.managed (Exception.mask_ . Async.withAsyncWithUnmask forwardOut))
-    c <- using (Managed.managed (Exception.mask_ . Async.withAsyncWithUnmask forwardErr))
+    a <- using
+        (Managed.managed (\k ->
+            Exception.mask (\restore ->
+                Async.withAsync (feedIn restore) k ) ))
+    b <- using
+        (Managed.managed (\k ->
+            Exception.mask (\restore ->
+                Async.withAsync (forwardOut restore) k ) ))
+    c <- using
+        (Managed.managed (\k ->
+            Exception.mask (\restore ->
+                Async.withAsync (forwardErr restore) k ) ))
     let l `also` r = do
             _ <- l <|> (r *> STM.retry)
             _ <- r
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -260,13 +260,13 @@ module Turtle.Prelude (
 import Control.Applicative
 import Control.Concurrent (threadDelay)
 import Control.Concurrent.Async
-    (Async, withAsync, withAsyncWithUnmask, waitSTM, concurrently,
+    (Async, withAsync, waitSTM, concurrently,
      Concurrently(..))
 import qualified Control.Concurrent.Async
 import Control.Concurrent.MVar (newMVar, modifyMVar_)
 import qualified Control.Concurrent.STM as STM
 import qualified Control.Concurrent.STM.TQueue as TQueue
-import Control.Exception (Exception, bracket, bracket_, finally, mask_, throwIO)
+import Control.Exception (Exception, bracket, bracket_, finally, mask, throwIO)
 import Control.Foldl (Fold, FoldM(..), genericLength, handles, list, premap)
 import qualified Control.Foldl
 import qualified Control.Foldl.Text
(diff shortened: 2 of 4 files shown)


mitchellwrosen:
Whoa... a PR assigned to me?!

Before digging in too deep I'm already confused by the old code:

```haskell
let feedIn :: (forall a. IO a -> IO a) -> IO ()
    feedIn restore =
        restore (ignoreSIGPIPE (outhandle hIn s)) `finally` close hIn
mask_ (withAsyncWithUnmask feedIn (\a -> Process.waitForProcess ph <* halt a) )
```

The `restore` function passed to `feedIn` would restore the masking state from `MaskedInterruptible` to... `MaskedInterruptible`, no?

Another question: although `waitForProcess` was running under `mask_`, surely this is a blocking operation to which async exceptions should be delivered?

Gabriel439:
I assigned it to you because you wrote the old code in 194cab4308cd2a1e501201d0ad97d52fbca31dea :)

To answer your question, the reason why `restore` works at all is that `withAsyncWithUnmask` lets you cheat and jail-break out of a `MaskedInterruptible` state into an `Unmasked` state.  See the documentation of the (similar) [`forkIOWithUnmask` function](https://hackage.haskell.org/package/base-4.9.1.0/docs/Control-Concurrent.html#v:forkIOWithUnmask).  I generally try to avoid using it and prefer using `mask` instead

It's not clear to me whether or not `waitForProcess` is treated as blocked by the runtime or not.

mitchellwrosen:
Ah! So I did. Your patch looks good! :+1:

One last question, you say

> The outhandle function running in the forked feedIn thread gets correctly
terminated by the exception because it runs unmasked inside of the restore

but how can this be, if the `ThreadKilled` is delivered to the main thread running `handle`, not its child running `feedIn`?

mitchellwrosen:
Hm, well after printf-debugging (the old version of the code) for some time, it appears that the call to `c_waitForProcess` is not being interrupted by a `ThreadKilled` as I'd expect. It's really a best-effort mechanism, GHC sends SIGPIPE to the thread and hopes it dies. In this case, the thread is just calling `waitpid` which does return EINTR on SIGPIPE. Sooo... not totally sure what's going on.

mitchellwrosen:
Curiously, a _single_ Ctrl+C (not two) is enough to kill the hanging program, which should be exactly the same mechanism that GHC uses for interrupting an interruptible foreign call, except the exception is `UserInterrupt`, not `ThreadKilled`. So why does Ctrl+C work, but `cancel` doesn't?

mitchellwrosen:
Some strace output:

```
-- Signal child thread 23987 with SIGPIPE
[pid 23989] tgkill(23982, 23987, SIGPIPE <unfinished ...>                                                                  
[pid 23989] <... tgkill resumed> )      = 0                            
                                                    
-- Thread 23987 receives SIGPIPE, should return EINTR
[pid 23987] <... wait4 resumed> 0x7f61c4a3ce6c, 0, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)            
[pid 23987] --- SIGPIPE {si_signo=SIGPIPE, si_code=SI_TKILL, si_pid=23982, si_uid=1000} ---                                
[pid 23987] rt_sigreturn({mask=[]} <unfinished ...>                                                                        
[pid 23987] <... rt_sigreturn resumed> ) = -1 EINTR (Interrupted system call)     

-- Yet it re-enters waitpid                                     
[pid 23987] wait4(23988,  <unfinished ...>                                                                                 
```

Gabriel439:
> One last question, you say
> 
> > The outhandle function running in the forked feedIn thread gets correctly
>  > terminated by the exception because it runs unmasked inside of the restore
> 
> but how can this be, if the ThreadKilled is delivered to the main thread running handle, not its child > running feedIn?

There are two separate `ThreadKilled` exceptions in play:

* The first `ThreadKilled` exception is the one that took down the thread running the sub-process (the one you are thinking of)
* The second `ThreadKilled` exception is triggered by the cleanup logic of the `fork` utility, which calls `cancel` on the child thread (i.e. `feedIn`)

So the `outhandle` function is terminated by the *second* `ThreadKilled` exception, whereas the `waitForProcess` is being protected from the *first* `ThreadKilled` exception due to not being unmasked

Also, I think I have another clue for why I was running into the hang.  The issue only seems to arise when interpreting Haskell code (i.e. using `ghci` or `runghc`).  In fact, the regression test I wrote was worthless because it actually always passes (even before the PR) if the program is compiled!

However, if you interpret the same regression test then it hangs (meaning that even the `System.Timeout.timeout` does not return correctly):

```bash
$ runghc test/RegressionMaskingException.hs 
<hangs>
$ ghc -O2 test/RegressionMaskingException.hs -o example
$ ./example  # Returns immediately
$ echo $?
0
```

Same behavior in `ghci` (even `System.Timeout.timeout` hangs)

Gabriel439:
Okay, I have a minimal reproducing test case using using the `process` library:

```haskell
$ ghci
>>> System.Timeout.timeout 1 (Control.Exception.mask_ (System.Process.system "while true; do sleep 1; done"))
<hangs>
>>> System.Timeout.timeout 1 (System.Process.system "while true; do sleep 1; done")
Nothing
```

Gabriel439:
Actually, this minimal example even fails when compiled:

```haskell
import Control.Exception
import System.Timeout
import System.Process

main :: IO ()
main = do
    _ <- timeout 1 (mask_ (system "while true; do sleep 1; done"))
    return ()
```

mitchellwrosen:
The discrepancies you were seeing with compiled vs. interpreted are due to the fact that the `system` call needs to actually get to `waitForProcess` before being thrown an async exception. Here's a slightly simpler version of your example:

```haskell
import Control.Concurrent                                                                                                  
import Control.Exception                                                                                                   
import System.Timeout                                                                                                      
import System.Process                                                                                                      
                                                                                                                           
main :: IO ()                                                                                                              
main = do                                                                                                                  
  tid <- forkIO $ do                                                                                                       
    mask_ (system "while true; do sleep 1; done")                                                                          
    pure ()                                                                                                                
                                                                                                                           
  threadDelay 1000 -- without this line, thread might be killed too early                                                                                                 
  throwTo tid ThreadKilled
```

I still haven't figured out why `waitForProcess` isn't interruptible!

mitchellwrosen:
Wrapping `waitForProcess` with `interruptible` seems to work.

```haskell
import Control.Concurrent                                                                                                  
import Control.Exception                                                                                                   
import System.Timeout                                                                                                      
import System.Process                                                                                                      
                                                                                                                           
main :: IO ()                                                                                                              
main = do                                                                                                                                                                                                                                  
  tid <- forkIO $ do                                                                                                                                                                                                                                 
    (_, _, _, h) <- createProcess (shell "sleep 5")                                                                        
    mask_ (interruptible (waitForProcess h))                                                                               
    return ()                                                                                                              
                                                                                                                           
  threadDelay 1000                                                                                                       
  throwTo tid ThreadKilled
```

Gabriel439:
Wrapping it with `interruptible` is equivalent to wrapping it with `restore` (assuming that the corresponding `mask` call was called from an unmasked context).  All `interruptible` does does it revert back to an unmasked state temporarily

It doesn't explain why `waitProcess` wasn't interruptible when it blocked.  So far, what I understand is that GHC makes a best effort attempt to terminate `interruptible` FFI calls by sending `SIGPIPE` according to [this](https://downloads.haskell.org/~ghc/master/users-guide/ffi-chap.html#interruptible-foreign-calls) but that wasn't enough in this case as your `strace` showed

Either way I will go ahead and merge this now that the regression test is fixed

mitchellwrosen:
Ah, I didn't look at the implementation of `interruptible`, I just assumed it was a primitive. I'm still not sure why the function is restarted as shown in the `strace` I pasted above. It seems sending `SIGPIPE` _was_ sufficient - the C call returned -1 with EINTR, rather than ignore the signal. Shrug!

PR #226, opened 2017-03-31
Add a helper function to get output from shell

NickSeagull:
This would allow to write things like

```haskell
main = do
    Just directory <- getShellOut "stack path --local-install-root"
    print directory
```

instead of 

```haskell
main = do
    Just directory <- fold (inshell "stack path --local-install-root" empty) Foldl.head
    print directory
```

I hope this helps in some way :smile:


Gabriel439:
I think this utility probably belongs in the `Turtle.Prelude` module, but I might propose restructuring it like this:

```haskell
single :: MonadIO a => Shell a -> io (Maybe a)
single s = do
    ls <- fold s Foldl.list
    case ls of
        [a] -> return (Just a)
        _   -> return Nothing
```

That way you can use it with any function that returns a `Shell` and it also fails if there is more than one line of output.

NickSeagull:
@Gabriel439 I just pushed the changes that you requested 😃

src/Turtle/Prelude.hs
```diff
@@ -1825,3 +1825,10 @@ header (Shell k) = Shell k'
             return (Pair x Nothing)
 
         done' (Pair x _) = done x
+
+single :: MonadIO a => Shell a -> io (Maybe a)
+single s = do
+    ls <- fold s Contorl.Foldl.list
```
Gabriel439: `s/Contorl/Control/`

NickSeagull:
I agree with those changes @Gabriel439 , it is true that the script should fail explicitly, because otherwise it is some unexpected behaviour. Just pushed those changes

src/Turtle/Prelude.hs
```diff
@@ -1825,3 +1825,12 @@ header (Shell k) = Shell k'
             return (Pair x Nothing)
 
         done' (Pair x _) = done x
+
+single :: MonadIO io => Shell a -> io (Maybe a)
```
Gabriel439: One last request: could you add documentation for this utility?

NickSeagull:
There we go! 😄

Gabriel439:
Thank you! :)

mitchellwrosen:
Just wondering, why does this return a `Maybe`?

Gabriel439:
Oops!  This should probably just return an `a` instead of a `Maybe a`

Gabriel439:
Type of `single` fixed in 71db10ab74b4f9aea75c2b116cca3c392d60c612

PR #233, opened 2017-04-26
Update whichAll to honor OS-specific search path delimiter

jship:
This PR replaces the hard-coded search path delimiter of `':'` used in `Turtle.Prelude.whichAll` with `system-filepath`'s `splitSearchPathString`.

This gets `which` and `whichAll` working on Windows.  On Windows, the search path delimiter is a semicolon from both git bash and cmd.exe so `which`/`whichAll` were returning empty.


Gabriel439:
Thank you for fixing this!  I really appreciate it :)

mitchellwrosen:
Whoops! I think I originally contributed `which`/`whichAll` and neglected to consider Windows paths.

PR #238, opened 2017-05-26
Integration test for cptree

strokyl:
I encounter the issue fixed by this PR: https://github.com/Gabriel439/Haskell-Turtle-Library/pull/235. (the issue still present on the version distributed by the latest, 8.15, lts of stack). I did not the commit that fix it, and I started wanted to fix the bug, I wrote this integration test that I could not success to make it fail because the bug was already fixed!

I do not know if you are interested in test but in case here my PR. I think it's good to have a test to prove bugfix work and to avoid regression.

### DIFF
--- a/test/cptree.hs
+++ b/test/cptree.hs
@@ -6,18 +6,16 @@ import System.IO.Temp (withSystemTempDirectory)
 import Control.Monad (unless)
 
 check :: String -> Bool-> IO ()
-check errorMessage successs = unless successs $ error errorMessage
+check errorMessage successs = unless successs $ fail errorMessage
 
--- This test fails by hanging
 main :: IO ()
 main = withSystemTempDirectory "tempDir" (runTest . fromString)
 
 runTest :: Turtle.FilePath -> IO ()
 runTest tempDir = do
   let srcDirectory = tempDir </> "src"
-  mkdir srcDirectory
 
-  mkdir $ srcDirectory </> "directory"
+  mktree $ srcDirectory </> "directory"
   touch $ srcDirectory </> "directory" </> "file"
 
   let destDirectory = tempDir </> "dest"
--- a/test/cptree.hs
+++ b/test/cptree.hs
@@ -1,0 +1,28 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+import Turtle
+import Filesystem.Path.CurrentOS ()
+import System.IO.Temp (withSystemTempDirectory)
+import Control.Monad (unless)
+
+check :: String -> Bool-> IO ()
+check errorMessage successs = unless successs $ error errorMessage
+
+-- This test fails by hanging
+main :: IO ()
+main = withSystemTempDirectory "tempDir" (runTest . fromString)
+
+runTest :: Turtle.FilePath -> IO ()
+runTest tempDir = do
+  let srcDirectory = tempDir </> "src"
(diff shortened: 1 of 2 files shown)


test/cptree.hs
```diff
@@ -0,0 +1,28 @@
...
+import Filesystem.Path.CurrentOS ()
+import System.IO.Temp (withSystemTempDirectory)
+import Control.Monad (unless)
+
+check :: String -> Bool-> IO ()
+check errorMessage successs = unless successs $ error errorMessage
+
+-- This test fails by hanging
+main :: IO ()
+main = withSystemTempDirectory "tempDir" (runTest . fromString)
```
Gabriel439: You can use `Turtle.tempdir` here:

```haskell
with (mktempdir "/tmp" "foo") :: (FilePath -> IO r) -> IO r
```

test/cptree.hs
```diff
@@ -0,0 +1,28 @@
...
+
+import Turtle
+import Filesystem.Path.CurrentOS ()
+import System.IO.Temp (withSystemTempDirectory)
+import Control.Monad (unless)
+
+check :: String -> Bool-> IO ()
+check errorMessage successs = unless successs $ error errorMessage
+
+-- This test fails by hanging
```
Gabriel439: If the test fails by hanging then this should use `System.Timeout.timeout` to detect the hang

test/cptree.hs
```diff
@@ -0,0 +1,28 @@
...
+main :: IO ()
+main = withSystemTempDirectory "tempDir" (runTest . fromString)
+
+runTest :: Turtle.FilePath -> IO ()
+runTest tempDir = do
+  let srcDirectory = tempDir </> "src"
+  mkdir srcDirectory
+
+  mkdir $ srcDirectory </> "directory"
```
Gabriel439: You can combine these two `mkdir` statements into one command with:

```haskell
mktree (srcDirectory </> "directory")

File: test/cptree.hs (hunk 2 above)
Comment (strokyl) [inline test/cptree.hs]: Sorry for this comment I start my test by copying pasting an other one and I forgot to remove the comment!

File: test/cptree.hs (hunk 1 above)
Comment (strokyl) [inline test/cptree.hs]: I saw Turtle.tempdir, but I prefer withSystemTempDirectory because it will use the correct system temp directory if it is not "/tmp":
```
On Unix, getTemporaryDirectory returns the value of the TMPDIR environment variable or "/tmp" if the variable isn't defined. On Windows, the function checks for the existence of environment variables in the following order and uses the first path found:

    TMP environment variable.
    TEMP environment variable.
    USERPROFILE environment variable.
    The Windows directory
```
https://hackage.haskell.org/package/directory-1.2.2.0/docs/System-Directory.html#v:getTemporaryDirectory

test/cptree.hs
strokyl: done

Gabriel439:
Looks great!  Thank you for doing this :)

strokyl:
You did not let me time to autosquash before merging :-(.
You welcome, I am looking for thinks to do in order to practice my Haskell and I found the idea of doing script in haskell awesome so if need some help for other test or issue I will be happy to help.

Gabriel439:
Don't worry.  There's no need for you to squash before merging.  GitHub lets me squash the changes when I merge

PR #241, opened 2017-06-11
Relaxing the upper bound of optparse-applicative to allow for the 0.14 release

ixmatus:
This change relaxes the upper bound of the `optparse-applicative` dependency to allow for the recent `0.14` release.

I built `turtle` with an overridden `optparse-applicative` derivation that provided the `0.14` version. The build succeeded without error.


Gabriel439:
Thank you :)

ixmatus:
Yup you're welcome :)

PR #248, opened 2017-07-15
`stream{,WithErr}` throw `ExitCode` on failure.  Fixes #245

### LINKED ISSUE #245: Throw IO exception in `stream`
Hello!

Is there a reason why `stream` ignores the `ExitCode` of the process? Since there is no room for a proper "return value" as in pipes, perhaps throwing an exception is preferable to doing nothing.

My use-case is logging the output of a process inside some larger script; if the sub-process fails, so should the whole script:

```haskell
runFoo :: Shell (Either Line Line)
runFoo = inshellWithErr "foo" empty

-- Flatten stdout and stderr
runFoo' :: Shell Line
runFoo' = fmap (either id id) runFoo

main :: IO ()
main = sh $ do
  output "logfile" runFoo'
  ...
  -- should not be reached if foo fails
  ...
```

Thanks!

Gabriel439:


Gabriel439:
@mitchellwrosen Good catch!  Thanks :)

Issue #251, opened 2017-08-17
MonadFail instance for Shell


matthewleon:
Should `Shell` adopt `MonadFail`?

Gabriel439:
Yes!  `Shell` already implements an explicit `fail` method for `Monad`, so it's only natural to use `MonadFail` now that it's available

PR #253, opened 2017-08-19
Shell MonadFail instance

matthewleon:
addresses https://github.com/Gabriel439/Haskell-Turtle-Library/issues/251

Uses some CPP. I don't see a way around that.

### DIFF
--- a/src/Turtle/Shell.hs
+++ b/src/Turtle/Shell.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RankNTypes, CPP #-}
 {-# OPTIONS_GHC -fno-warn-missing-methods #-}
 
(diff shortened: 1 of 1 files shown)


src/Turtle/Shell.hs
```diff
@@ -151,6 +154,11 @@ instance MonadManaged Shell where
         x' <- with resource (step x)
         done x' )
 
+#if MIN_VERSION_base(4,9,0)
+instance Fail.MonadFail Shell where
+    fail = fail
```
Gabriel439: Could you qualify the right-hand side as `Prelude.fail` just so that we don't run the risk of introducing an infinite loop in the future if we import `Control.Monad.Fail.fail` unqualified?

matthewleon: Good catch.

matthewleon:
Should be ready for review again.

Gabriel439:
Thanks!  This looks great and I'll merge once CI completes

PR #258, opened 2017-09-24
Update tutorial

edkolev:
Replace "A simple `cp` script" with "A simple `cp` utility" to match example output


edkolev:
This is a slight change to the tutorial to address the following issue:

- example code is:
```
main = do
    (src, dest) <- options "A simple `cp` script" parser
    cp src dest
```

- the example output is:
```
$ ./cp.hs --help
A simple `cp` utility
```
Hence this PR to replace `script` => `utulity`

Gabriel439:
Thank you for fixing that! :)

PR #260, opened 2017-11-08
Add `grepText` to the Prelude

markus1189:
During my scripting, I missed a function that could `grep` a `Shell Text` and so I wrote one ;)

I added an internal helper `grepWith` that is parameterized over an additional function wich converts to `Text` and used that one from both `grep` and `grepText`.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -171,6 +171,7 @@ module Turtle.Prelude (
     , lstree
     , cat
     , grep
+    , grepText
     , sed
     , sedPrefix
     , sedSuffix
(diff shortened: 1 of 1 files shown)


src/Turtle/Prelude.hs
```diff
@@ -1513,12 +1514,16 @@ appendonly file = using (managed (Filesystem.withTextFile file IO.AppendMode))
 cat :: [Shell a] -> Shell a
 cat = msum
 
+grepWith :: (b -> Text) -> Pattern a -> Shell b -> Shell b
+grepWith f pattern s = mfilter (not . null . match pattern . f) s
```
markus1189: Was there a reason to *not* use `mfilter` before this?

Gabriel439: No reason.  `mfilter` is just fine :)

Gabriel439:
I'll merge this as soon as CI passes

markus1189:
We could also add a `grepFilePath :: Pattern a -> Shell FilePath -> Shell FilePath` which would be just `onFiles . grepText` but I don't know whether it is worth to make that more visible or if users should just write the version with `onFiles`

Gabriel439:
Yeah, I would recommend using `onFiles` for the `FilePath` case

Gabriel439:
I can push a new release to Hackage if you have no other pending changes

markus1189:
You are really super fast, I appreciate that.  Let's wait one or two days, I might find more stuff to add and I don't mind having it defined inside my script for now ;)

markus1189:
You are really super fast, I appreciate that.  Let's wait one or two days, I might find more stuff to add and I don't mind having it defined inside my script for now ;)

markus1189:
You are really super fast, I appreciate that.  Let's wait one or two days, I might find more stuff to add and I don't mind having it defined inside my script for now ;)

markus1189:
You are really super fast, I appreciate that.  Let's wait one or two days, I might find more stuff to add and I don't mind having it defined inside my script for now ;)

PR #261, opened 2017-11-09
Add `uniq`, `nub` and `sorted` to the `Prelude`

markus1189:
I had to add `containers` to the dependencies, but I guess that is not a problem anyway.

- uniq: basically the same as the `uniq` commandline tool, discards adjacent duplicates
- nub: does a global removal of duplicates, at the cost of having a `Set` in memory.  We could improve on this by using a `hash`, but I think adding that would be a lot more changes, although it *is* better ;)
- sorted: This is not a function from and to `Shell`, because we have to run it to the end and then do the sorting.  I kept it the same as the commandline `sort`, so duplicates are *not* removed

The whole `IO` + `IORef` thing feels a little dirty though :D

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE TupleSections              #-}
 
 -- | This module provides a large suite of utilities that resemble Unix
 --  utilities.
@@ -191,6 +192,14 @@ module Turtle.Prelude (
     , cache
     , parallel
     , single
+    , uniq
+    , uniqOn
+    , uniqBy
+    , nub
+    , nubOn
+    , sort
+    , sortOn
+    , sortBy
 
     -- * Folds
     , countChars
@@ -280,7 +289,7 @@ import Control.Concurrent.MVar (newMVar, modifyMVar_)
 import qualified Control.Concurrent.STM as STM
 import qualified Control.Concurrent.STM.TQueue as TQueue
 import Control.Exception (Exception, bracket, bracket_, finally, mask, throwIO)
-import Control.Foldl (Fold, FoldM(..), genericLength, handles, list, premap)
+import Control.Foldl (Fold(..), FoldM(..), genericLength, handles, list, premap)
 import qualified Control.Foldl
 import qualified Control.Foldl.Text
 import Control.Monad (guard, liftM, msum, when, unless, (>=>), mfilter)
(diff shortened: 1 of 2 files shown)


src/Turtle/Prelude.hs
```diff
@@ -1986,3 +1991,45 @@ single s = do
...
+-- | Filter adjacent duplicate elements:
+--
+-- >>> view (uniq (select [1,1,2,1,3]))
+-- 1
+-- 2
+-- 1
+-- 3
+uniq :: Eq a => Shell a -> Shell a
+uniq s = Shell $ \(FoldM step begin done) -> do
+  ref <- newIORef Nothing
```
Gabriel439: I believe you can do this without an `IORef`.  I haven't type-checked this, but I believe it would go something like this:

```haskell
uniq s = Shell $ \(FoldM step begin done) -> do
    let step' (x, Just a') a | a == a' = do
            return (x, Just a)
        step' (x, _      ) a | otherwise = do
            x' <- step x a
            return (x', Just a)

    let begin' = (begin, Nothing)

    let done' (x, _) = done x

    foldIO s (FoldM step' begin' done')
```

This takes advantage of the fact that the existentially quantified state of the new `FoldM` doesn't have to have the same type as the existentially quantified state of the fold `FoldM`

Similar comment for `nub`

3noch:
Consider adding `By` variants.

src/Turtle/Prelude.hs
markus1189: Yeah that is *so* much better, or at least I can sleep better

markus1189:
Added the `-By` variants and removed the `IORef`s ;)

int-index:
> Added the -By

I think what you added are `-On` variants (compare `sortBy` and `sortOn` from `base`)

markus1189:
Yeah good catch.  Somehow I like the `on` variants better.  Maybe just rename it to `-On`?

int-index:
Admittedly, I need `sortOn` more often than `sortBy`. But the `-By` variant is more general. Why not both?

markus1189:
Okay I added both `-On` and `-By`.  The only variant that is not present is `nubBy`, because here we need the `Ord` instance for the `Set`, so we only have `nubOn`.

Is the distinction and naming of `nub` vs `uniq` okay?  I am unsure currently...

Gabriel439:
I think the `uniq` and `nub` naming convention is good

If nobody has any other objections I'll merge this tomorrow

mitchellwrosen:
I object! Lightly! To `sorted`. Why `sorted` and not `sort`? We already clash with `nub` in this patch.

Also, it doesn't really _do much_, but I get that shell scripting should be nice and terse. Sort of at odds with my preference for a minimal API.

markus1189:
I'm also okay with sort, we can change that.

And yeah I was also hesitant to add it, but it mirrors the command line
utility and I guess it is nicer to provide it for less experienced
turtlers. Although I don't have a strong opinion about that, I wouldn't
mind removing it again

markus1189:
Okay renamed to `sort`, I agree to what @mitchellwrosen says ;)

Gabriel439:
I will push a new release to Hackage if nobody objects because I need to increase the upper bound on `unix-compat` for Stackage anyway

Issue #209, opened 2017-01-31
It's difficult to work with functions that expect lazy ByteString.


crocket:
Since `Shell ByteString` emits chunks of strict `ByteString`, it's difficult to marshal between `Turtle.Bytes` and functions expecting lazy `ByteString`.

I found it easier to ditch `turtle` and use `optparse-applicative` directly as below when I use functions expecting lazy `ByteString`.

```haskell
#!/usr/bin/env stack
{- stack --install-ghc runghc
   --package optparse-applicative
   --package charsetdetect-ae
   --package iconv
-}
{-# LANGUAGE OverloadedStrings #-}
import qualified Data.ByteString.Lazy.Char8 as C
import Codec.Text.Detect (detectEncodingName)
import Codec.Text.IConv (convert)
import Options.Applicative

optParser :: Parser FilePath
optParser = strArgument (metavar "FILE")

main :: IO ()
main = do
  let desc = "Converts the encoding of a file's content to UTF-8, "
        ++ "and remove carriage returns."
  file <- execParser $ info (helper <*> optParser) (progDesc desc)
  bs <- C.readFile file
  case detectEncodingName bs of
    Nothing -> putStrLn "Charcter encoding detection failed"
    Just "UTF-8" -> putStrLn "Character encoding is already UTF-8"
    Just ce -> do
      putStrLn $ "Detected character encoding : " ++ ce
      C.writeFile file $ C.filter (not . (=='\r')) $ convert ce "UTF-8" bs
```

In the above example, `detectEncodingName` and `convert` expect lazy `ByteString`.
If I used `Turtle.Bytes.input` and `Turtle.Bytes.output` instead of `C.readFile` and `C.writeFile`, verbose marshalling would be necessary.
I think `Turtle.Bytes` needs lazy variants.

Gabriel439:
One solution would be to add a `Turtle.Prelude.lazy` and `Turtle.Bytes.lazy` which are identical to `Turtle.Prelude.strict` and `Turtle.Bytes.strict` except they return lazy `Text`/`ByteString`.  Would that solve your issue?

crocket:
I have two questions.

1. How is lazy `Text` in `Turtle.Prelude.lazy` going to help?
2. Is there a use case for chunks of strict `ByteString` emitted by functions in `Turtle.Bytes`?

Gabriel439:
[...]

PR #267, opened 2017-11-21
Add ByteString.Lazy utilities

berewt:
The starting point was a similar issue from the one raised in #209. Compared to the solution proposed in the thread, I'd prefer to go for a whole set of utilities for `ByteString.Lazy`. It allows almost the same shell facilities than the strict version, except that chunks are meaningless here.


Gabriel439:
Wow, thanks for doing this!  Give me a day or two to review this more closely, but I approve of the overall direction of this

src/Turtle/Bytes/Lazy.hs
```diff
@@ -0,0 +1,639 @@
...
+import qualified Data.ByteString.Lazy
+import qualified Data.Text
+import qualified System.IO
+import qualified System.Process                as Process
+import qualified Turtle.Prelude
+
+{-| Read lazily bytes from standard input
+
+-}
+stdin :: Shell ByteString
```
Gabriel439: Should this have type:

```haskell
stdin :: MonadIO io => io ByteString
```

src/Turtle/Bytes/Lazy.hs
```diff
@@ -0,0 +1,639 @@
...
+stdin = inhandle System.IO.stdin
+
+{-| Read chunks of bytes from a file
+
+    The chunks are not necessarily aligned to line boundaries
+-}
+input :: FilePath -> Shell ByteString
```
Gabriel439: Should this have type:

```haskell
input :: MonadManaged managed => FilePath -> managed ByteString
```

src/Turtle/Bytes/Lazy.hs
```diff
@@ -0,0 +1,639 @@
...
+        ExitSuccess   -> return ()
+        ExitFailure _ -> Exception.throwIO exitCode
+
+{-| `stream` generalizes `inproc` and `inshell` by allowing you to supply your
+    own custom `CreateProcess`.  This is for advanced users who feel comfortable
+    using the lower-level @process@ API
+
+    Throws an `ExitCode` exception if the command returns a non-zero exit code
+-}
+stream
```
Gabriel439: I don't think you need any of the streaming functions like `stream`/`streamWithErr` or the functions downstream of them.  The `systemStrict` and `systemStrictWithErr` versions already stream since they are using `Data.ByteString.Lazy.hGetContents`

berewt:
Regarding the signature changes: yep, I forgot to transitively generalise the functions that depends on `inhandle`, thanks.

Regarding the `stream` / `streamWithErr`, maybe I should only keep the strict version and rename them as `stream` and `streamWithErr`, as the `strict` names can be misleading, what do you think about it?

Gabriel439:
Yeah, I agree that they don't need the `strict` in the names for this module

berewt:
One more question then, if we remove `stream` / `streamWithError` in favour of `systemStrict` / `systemStrictWithError`, what do we do with function like `inProc`, that was previously relying on `stream`, and will now use `systemStrict`:

1. keep their signature as is and discard the `ExitCode`;
2. change them, to include the `ExitCode`?

Gabriel439:
I think you don't even `inproc` any longer since `proc` and the new `stream` utility supersede it

Or you could rename the new `stream` utility to `inproc` instead for conceptual consistency

berewt:
I finally decided to drop the `in*` function and went for a `*Stream` name instead. The names are consistent with the `*Strict` equivalent in `Turtle.ByteString`. Please do not hesitate to propose something else if you're not happy with the result, I'm not totally satisfied myself either. ;-)

src/Turtle/Bytes/Lazy.hs
```diff
@@ -0,0 +1,422 @@
...
+    :: MonadIO io
+    => Process.CreateProcess
+    -- ^ Command
+    -> Shell ByteString
+    -- ^ Chunks of bytes written to process input
+    -> io (ExitCode, ByteString)
```
Gabriel439: Actually, something seems wrong here.  This type implies that we have access to the `ExitCode` as the same time as the `ByteString`, which is being read lazily.  If we are able to access the `ExitCode`, wouldn't that imply that the reading `ByteString` would fail due to the subprocess handle being closed?

berewt: I guess that I should have port the initial stream version as well then... I should really try and test these functions with some examples, I'm not sure whether the out handle will be accessible or not in that case.

Gabriel439: One option would be to read the lazy `ByteString` strictly (i.e. force the entire `ByteString`) so that you don't have to worry about still reading it after the handle is closed

PR #270, opened 2017-12-10
Allow 'err' and 'echo' to take in multiline text

lynnard:
The `Line` type enforces a semantics that disallow having newlines in the input, but personally I find it a valid use case to `echo` or `err` text with newlines in it - in which case the defined behavior is to print the text and then append a newline in the end. As long as this 'defined behavior' is sufficiently intuitive / sensible to users, we shouldn't need to restrict the acceptable input - what do you think?

### DIFF
--- a/src/Turtle/Format.hs
+++ b/src/Turtle/Format.hs
@@ -46,6 +46,7 @@ module Turtle.Format (
     , (%)
     , format
     , printf
+    , eprintf
     , makeFormat
 
     -- * Parameters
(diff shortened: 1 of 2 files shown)


int-index:
I don't know if we should do this, what would be the preferred way to print a `Line` after this change? Using `lineToText` manually would be a bit too verbose.

So I'm not sure we should make this change. Even if we do, please also update the documentation, as Haddock for `echo` and `err` still say that they are for printing exactly one line (not true after your changes).

lynnard:
Maybe the problem is with the clunkiness for conversion between `Text` and `Line`. I agree `lineToText` is a bit verbose, but compare with currently converting from `Text` to `Line`:

```
mapM_ err . textToLines $ ...
```

int-index:
What about adding `eprintf` similar to `printf` but for stderr?

Gabriel439:
Yeah, I was going to suggest `printf s`, so I think `eprintf` would be a good addition

lynnard:
Closing this in favor of #271

PR #271, opened 2017-12-12
add 'eprintf' to faciliate printing to stderr

lynnard:


Gabriel439:
Thank you for adding this!

PierreR:
Is it really a good idea to use such short name ? Why not errPrintf ? 
The protolude is using putErrText for instance.

Gabriel439:
@PierreR: I thought `eprintf` was the idiomatic name that people usually used for `printf` to `stderr`

PierreR:
If it is let's be it ;-)

PR #273, opened 2017-12-16
Document `inproc`: throws `ExitCode` for non-zero exit codes

damiencourousse:
hello, 

I think it is worth mentionning that `inproc` throws `ExitCode`, because without reading the source code, it is not obvious that `inproc` is based on `stream`.

Thank you for this nice library, BTW!
Damien


Gabriel439:
You're welcome and thank you for the pull request! :)

Issue #278, opened 2018-01-05
Stackage `lts-10` and `Win32 < 2.4`


sgraf812:
Appearently, `turtle-1.4.5` made it into Stackage LTS 10 (yay?!), which includes `Win32-2.5.4.1` in its package set. Unfortunately, that seems to be incompatible with turtle. This probably slipped CI because there isn't really a good free CI story for Windows.

Gabriel439:
No problem.  I will try to fix this over the weekend.  I have a local Windows machine I can use for testing `turtle` on Windows

Gabriel439:
Closing, since I assume that #279 fixes this issue.  Feel free to reopen if there is still more work to do

PR #279, opened 2018-01-05
Loosened upper bound for Win32

sgraf812:
In response to #278. The same tests seem to break with and without this change (`regression-broken-pipe` + `regression-masking-exception`) for `stack test --fast --stack-yaml=stack-lts-10.yaml`.


Gabriel439:
Whoops!  Sorry that I missed your pull request.  Thank you for doing this

PR #283, opened 2018-02-01
Fix doc comment for single

jhrcek:
@Gabriel439 thank you for this wonderful library! I feel so invincible when refactoring my cli scripts based on this library - as opposed to working with bash scripts :) 

This is a minor doc fix related to https://github.com/Gabriel439/Haskell-Turtle-Library/commit/71db10ab74b4f9aea75c2b116cca3c392d60c612


Gabriel439:
You're welcome! :)

Also, great catch and fix

Issue #298, opened 2018-03-24
Feature proposal: gfind


cfhammill:
I think having a more full featured find utility would be great. 

I figured I'd post a plan to get ideas before I start trying to write it. In my head it's `gfind` because it's going to implement more of the features of gnu find.

## Desired Features (a start at)

- Non-recursive use
- Depth limitation like `-maxdepth` and `-mindepth`
- File status limitation like `-cmin` and `-amin`
- File type limitation like `-type`

## Implementation concerns

A feature rich find would have a huge number of parameters most of which wouldn't be necessary for every use. So optional arguments are almost definitely necessary. I think an approach inspired by: http://neilmitchell.blogspot.ca/2008/04/optional-parameters-in-haskell.html could be useful.

To avoid namespacing issues we could use [`vinyl`](https://hackage.haskell.org/package/vinyl) records, an example might look like:

```
gfind path (chars) ((MinDepth ==: 1) :& (MaxDepth == : 1))
```

A global set of defaults can be defined by the package and the users options can be integrated with the defaults via [`rreplace`](https://hackage.haskell.org/package/vinyl-0.8.1.1/docs/Data-Vinyl-Lens.html#v:rreplace)

Any thoughts and suggestions would be appreciated.

Gabriel439:
My intution is that if we design this correctly we won't need to have many parameters.  Some of these desired features could be split out into orthogonal utilities.  In the spirit of the Unix philosophy we want each utility to do one thing and do it well!

Specifically, I think you could split it up like this:

*   Create a `lsdepth` utility that is like `lstree` except with a minimum and maximum depth

    ```haskell
    lsdepth :: Int -> Int -> FilePath -> Shell FilePath
    ```

*   Structure the generalized `find` to take a `Shell FilePath` as an input instead of a `FilePath`

    ```haskell
    findtree :: Pattern a -> Shell FilePath -> Shell FilePath
    ```

    This way users can customize what files to traverse by supplying `ls`, `lstree`, `lsdepth`, or their own hand-written directory traversal:

    ```haskell
    findtree (suffix ".exe") (ls ".") -- Only search the current directory

    findtree (suffix ".exe") (lsdepth 0 3 ".") -- Only search three levels down
    ```

*   For file status filter, provide simple predicate functions, like:

    ```haskell
    cmin, cmax :: UTCTime -> FilePath -> Bool
    ```

    Then filtering the returned files just becomes `Control.Monad.mfilter`

[...]

PR #300, opened 2018-03-27
Add lsdepth, findtree, cmin, and cmax

cfhammill:
As discussed in #298 

Only issue here is that `cmin` and `cmax` need to be done in a `MonadIO`, so as far as I can tell it can't be used with `mfilter` with them directly.

Code review would be appreciated, I'm still pretty new to haskell.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -170,6 +170,7 @@ module Turtle.Prelude (
     , ls
     , lsif
     , lstree
+    , lsdepth
     , cat
     , grep
     , grepText
@@ -183,6 +184,7 @@ module Turtle.Prelude (
     , inplaceSuffix
     , inplaceEntire
     , find
+    , findtree
     , yes
     , nl
     , paste
@@ -269,6 +271,8 @@ module Turtle.Prelude (
     , PosixCompat.isDirectory
     , PosixCompat.isSymbolicLink
     , PosixCompat.isSocket
+    , cmin
+    , cmax
 
     -- * Headers
     , WithHeader(..)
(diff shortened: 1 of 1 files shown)


src/Turtle/Prelude.hs
```diff
@@ -1635,6 +1662,34 @@ find pattern dir = do
...
+  path <- files
+  Right txt <- return (Filesystem.toText path)
+  _:_ <- return (match pat txt)
+  return path
+
+{- | Check if a file was last modified after a given
+     timestamp
+-}     
+cmin :: MonadIO io => UTCTime -> FilePath -> io Bool
+cmin t file =
```
Gabriel439: Minor stylistic suggestion: I'd suggest using a non-pointfree function:

```haskell
cmin t file = adapt <$> lstat file
  where
    adapt x = posixSecondsToUTCTime (modificationTime x) > t
```

src/Turtle/Prelude.hs
```diff
@@ -1635,6 +1662,32 @@ find pattern dir = do
...
+  adapt <$> lstat file
+  where
+    adapt x = posixSecondsToUTCTime (modificationTime x) > t
+
+{- | Check if a file was last modified before a given
+     timestamp
+-}     
+cmax :: MonadIO io => UTCTime -> FilePath -> io Bool
+cmax t file =
+  adapt <$> lstat file
```
Gabriel439: It looks like Travis is failing on older versions of GHC due to `Functor` not being a super-class of `MonadIO`.  You can fix that by using `liftM` or `do` notation

cfhammill:
Hopefully this is fixed now, the build failed due to the hackage outage.

PR #307, opened 2018-06-07
Add symlinking for non-windows systems.

cfhammill:


cfhammill:
yeah, that's a much better idea, I'll switch to using that. Do you think my mingw check is sufficient for checking if symlinking is appropriate?

Gabriel439:
@cfhammill: Yeah, I think that check is good enough for determining if the system is Unix or not

cfhammill:
should be fixed now!

PR #312, opened 2018-07-03
Refactor Nix code

Gabriel439:
This pins `nixpkgs` and also uses `readDirectory` for ease of testing changes
to dependencies


miguel-negrao:
Is optparse-applicative.nix still needed ? I get the feeling it's not used anymore in the overrides.

PR #337, opened 2018-12-09
fixes possible chmod bug

atpotts:
Apologies for such a low value PR, and if I have misunderstood the intentions of any of the functions.

I was trying to use the chmod function recently and noticed that it was failing to change the permissions of any paths. 

This seems to be because it is calculating the new permissions and then writing back the old ones. This changes it to write back the new ones, which gives it behaviour closer to what I'd expect from reading the documentation.


Gabriel439:
No apology necessary!  Thank you for fixing this bug :slightly_smiling_face:

PR #344, opened 2019-02-06
Fix `cptree` to handle symlinks correctly

### LINKED ISSUE #343: cptree doesn't check for symlink loops, loops forever
Hi,

imagine the following directory structure:
```
mkdir /tmp/foo
ln -s /tmp/foo /tmp/foo/foo
```
Now when you run `cptree "/tmp/foo" "/tmp/bar"`, it treats the symlink as a directory and recurses, and you end up with a directory structure like `/tmp/bar/foo/foo/foo/foo/foo/foo/...`.

So before recursing it should probably check whether the file is a symlink and if so, create a new symlink in the target directory.

Note: I haven't checked how rmtree behaves, but if it does the same (i. e. recurse into symlinked directories), the results may be catastrophic. Just imagine trying to delete a symlink to `$HOME` :scream:

Gabriel439:
Fixes #343

If `cptree` encounters a symlink, just copy the symlink instead of copying
the file it points to or descending into any directory it points to.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -130,6 +130,7 @@ module Turtle.Prelude (
 #if !defined(mingw32_HOST_OS)
     , symlink
 #endif
+    , isNotSymbolicLink
     , rm
     , rmdir
     , rmtree
@@ -1095,12 +1096,18 @@ symlink a b = liftIO $ createSymbolicLink (fp2fp a) (fp2fp b)
   
 #endif
 
+{-| Returns `True` if the given `FilePath` is not a symbolic link
+
+    This comes in handy in conjunction with `lsif`:
+
+    > lsif isNotSymbolicLink
+-}
+isNotSymbolicLink :: MonadIO io => FilePath -> io Bool
+isNotSymbolicLink = fmap (not . PosixCompat.isSymbolicLink) . lstat
+
 -- | Copy a directory tree
 cptree :: MonadIO io => FilePath -> FilePath -> io ()
 cptree oldTree newTree = sh (do
-    let isNotSymbolicLink =
-          fmap (not . PosixCompat.isSymbolicLink) . lstat
-
     oldPath <- lsif isNotSymbolicLink oldTree
 
     -- The `system-filepath` library treats a path like "/tmp" as a file and not
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -1098,17 +1098,39 @@ symlink a b = liftIO $ createSymbolicLink (fp2fp a) (fp2fp b)
 -- | Copy a directory tree
 cptree :: MonadIO io => FilePath -> FilePath -> io ()
 cptree oldTree newTree = sh (do
-    oldPath <- lstree oldTree
+    let isNotSymbolicLink path = do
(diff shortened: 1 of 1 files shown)


src/Turtle/Prelude.hs
```diff
@@ -1098,17 +1098,39 @@ symlink a b = liftIO $ createSymbolicLink (fp2fp a) (fp2fp b)
 -- | Copy a directory tree
 cptree :: MonadIO io => FilePath -> FilePath -> io ()
 cptree oldTree newTree = sh (do
-    oldPath <- lstree oldTree
+    let isNotSymbolicLink path = do
```
mberndt123: How about
```
let isNotSymbolicLink = fmap (not . PosixCompat.isSymbolicLink) . lstat
```

src/Turtle/Prelude.hs
```diff
@@ -1098,17 +1098,39 @@ symlink a b = liftIO $ createSymbolicLink (fp2fp a) (fp2fp b)
...
+            return (not (PosixCompat.isSymbolicLink fileStatus))
+
+    oldPath <- lsif isNotSymbolicLink oldTree
+
     -- The `system-filepath` library treats a path like "/tmp" as a file and not
     -- a directory and fails to strip it as a prefix from `/tmp/foo`.  Adding
     -- `(</> "")` to the end of the path makes clear that the path is a
     -- directory
     Just suffix <- return (Filesystem.stripPrefix (oldTree </> "") oldPath)
+
```
mberndt123: Why not
```
let Just suffix = Filesystem.stripPrefix (oldTree </> "") oldPath
```

Gabriel439: @mberndt123: They are not equivalent.  The reason why is that Haskell interprets an irrefutable pattern match in a `let` binding differently than when binding a value in a `do` block.

An irrefutable pattern match in a `let` binding:

```haskell
let Just x = foo
```

... is the same as:

```haskell
let x = case foo of
        Just y -> y
        Nothing -> error "…: Irrefutable pattern failed for pattern Just x"
```

... whereas an irrefutable pattern match in the bind of a `do` block:

```haskell
Just x <- return foo
```

... is the same as:

```haskell
r <- return foo

x <- case r of
    Just y  -> return y
    Nothing -> fail "Pattern match failure in do expression at …"
```

These differ because `fail` might not be the same as `error` (in fact, it's very common for `fail` to not be `error`).  For example, for any list-like `Monad` (including `Shell`), `fail _ = empty` so an irrefutable pattern match when binding a value in a `do` block actually gracefully degrades to returning a total (albeit empty) result, whereas an irrefutable pattern match in a `let` binding gives a partial result (i.e. `error`) which is not intended to be recoverable.

So that's why that codde does the whole dance of wrapping the value in `return` before attempting an irrefutable pattern match.  Another way it could have been written is to explicitly desugar it, like this:

```haskell
x <- case Filesystem.stripPrefix (oldTree </> "") oldPath of
    Nothing -> empty
    Just x  -> return x
```

mberndt123: I see, thanks for explaining!

Gabriel439: You're welcome!

Issue #340, opened 2018-12-24
Support time 1.9


crocket:
I can't use time 1.9 on programs that use turtle, yet.

PR #352, opened 2019-08-26
GHC 8.8 MonadFail preparedness.

quasicomputational:
Always provides a `MonadFail Shell` instance, avoiding the evils of
conditional APIs.

Only define `Monad.fail` conditional on base being older than 4.13,
since it goes away in that version.


quasicomputational:
This should be the only code change needed for GHC 8.8 - bounds will still need to be bumped (e.g., #340), and of course dependencies will need to catch up.

PR #355, opened 2019-09-12
Added toLines.

### LINKED ISSUE #354: Concatenate neighboring Texts in the stream until a newline is reached
A convenient way to convert misaligned lines within a stream. Continued from #353.

GregorySchwartz:
Resolves #354. Probably needs a better name, should avoid partial functions, and needs verification if bang patterns are needed, especially on `stream`.

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes                 #-}
 {-# LANGUAGE TupleSections              #-}
+{-# LANGUAGE BangPatterns               #-}
 
 -- | This module provides a large suite of utilities that resemble Unix
 --  utilities.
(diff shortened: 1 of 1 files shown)


GregorySchwartz:
For educational purposes, is the constant space due to the strict pair?

Gabriel439:
@GregorySchwartz: The main difference is that in the streaming version the outer `step'` function immediately passes each complete `Line` to the inner `step` function here:

```haskell
-- Emit each complete `Line`
x' <- foldM step x (NonEmpty.init lines')
```

... before moving on to the next `Text` chunk.  In particular, the streaming version does not store the complete sequence of `Line`s inside of the accumulator as a `Shell Line`.

Passing completed `Line`s to the inner `step` function immediately  instead of holding onto them ensures that even if there are a large (or infinite) number of output `Line`s then the `toLine`s function still makes progress and doesn't wait until computing all of the `Line`s to begin forwarding them downstream.

Gabriel439:
I'll go ahead and merge #356, then.  However, I'd still like you to review #357 if you can to see if it addresses your use case before merging that one

PR #357, opened 2019-09-13
Add `Turtle.Bytes.toUTF8`

### LINKED ISSUE #353: Feature request: gzip
All streaming operations are done on files it looks like. There's no way to handle a compressed stream like `cat file.gz | gzip -d -c | head` or the like. This is possible with http://hackage.haskell.org/package/streaming-commons-0.2.1.1/docs/Data-Streaming-Zlib.html or there can be a bytestring `input` that can be decompressed and converted to `Shell Line`.

Gabriel439:
Fixes #353

### DIFF
--- a/src/Turtle/Bytes.hs
+++ b/src/Turtle/Bytes.hs
@@ -16,6 +16,7 @@ module Turtle.Bytes (
     , append
     , stderr
     , strict
+    , toUTF8
 
     -- * Subprocess management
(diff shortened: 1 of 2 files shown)


GregorySchwartz:
@Gabriel439 I think there should be a `fromUTF8` for symmetry (and for decompress -> work -> compress).

PR #358, opened 2019-09-16
Add `Turtle.Bytes.{compress,decompress}`

Gabriel439:
Fixes https://github.com/Gabriel439/Haskell-Turtle-Library/issues/353

### DIFF
--- a/nix/turtle.nix
+++ b/nix/turtle.nix
@@ -8,7 +8,7 @@
 mkDerivation {
   pname = "turtle";
   version = "1.5.14";
-  src = ./.;
+  src = ./..;
   libraryHaskellDepends = [
     ansi-wl-pprint async base bytestring clock containers directory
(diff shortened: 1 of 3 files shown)


GregorySchwartz:
This will fail with `gzip` archives as the window bits are "15 is for zlib compression, and 31 for gzip compression", and a lower value (15) than the archive value (31 for gzip) may result in `ZlibException (-3)` or `Z_DATA_ERROR`. The solution would be to change `decompress` and `compress` functions from `Shell ByteString -> Shell ByteString` to `WindowBits -> Shell ByteString -> ShellByteString`, i.e. allow the user to specify the `WindowBits`.

GregorySchwartz:
This seems to work well now!

Gabriel439:
@GregorySchwartz: Great!  Then I'll go ahead and merge

Issue #380, opened 2020-03-16
Possible memory leak


GregorySchwartz:
Not sure where, but this has a stack overflow without optimizations, and with optimizations it grows in memory:

`reduce Fold.length . toLines . TB.toUTF8 . TB.decompress (WindowBits 31) . TB.input $ "file.txt.gz"`

I would think it is streaming and should just need constant memory, right? Looking at the profiling, it looks like the issue is between `toLines . TB.toUTF8 . TB.decompress (WindowBits 31)`.

GregorySchwartz:
Attached is a partial heap profile for the above code.
![test](https://user-images.githubusercontent.com/2490088/76799893-d55c8b00-67a8-11ea-9d38-00fe520cd94c.png)

GregorySchwartz:
Actually, it happens with just `reduce Fold.length . input $ "file.txt"`. Am I misunderstanding the streaming / strict fold of `foldl` and `turtle`?

GregorySchwartz:
If I use `Fold.fold` on `Fold.fold Fold.minimum . lines <$> readFile "test.tsv"` there seems to be no real issue (not a solid horizontal line but a jagged one, but it's still horizontal so no real growth). Is there an issue with `Shell`'s `fold` or `reduce`? Maybe it's instance?

GregorySchwartz:
Also, `reduce Fold.minimum $ select [1..1000000000]` has a linear increase in memory use, while `Fold.fold Fold.minimum [1..1000000000]` does not. It must be a `Shell` issue.

Gabriel439:
@GregorySchwartz: Yeah, `Turtle`'s `fold` should run in constant space for these examples.  I'm still looking into this

GregorySchwartz:
@Gabriel439 Is there anything I can do to help with this issue? `Turtle` has a critical role in my programs and this high memory usage / potential slowness will affect benchmarks. I'm not too familiar with the internals of `Shell` is the only problem...

Gabriel439:
@GregorySchwartz: I think I know what the problem is.  The big clue is that `turtle` leaks but `foldl` does not, and I suspect the reason why is because `turtle`'s `fold` uses `translate` under the hood which introduces a space leak:

```haskell
translate :: FoldM IO a b -> FoldShell a b
translate (FoldM step begin done) = FoldShell step' Nothing done'
  where
    step' Nothing a = do
        x  <- begin
        x' <- step x a
        return (Just x')
[...]

PR #381, opened 2020-03-28
Fix space leak in `Shell`s

Gabriel439:
Fixes https://github.com/Gabriel439/Haskell-Turtle-Library/issues/380

The issue was that all of the `Shell` utilities were internally using
`translate` which was inadvertently using a non-strict `Maybe`
accumulator.  Switching to a strict accumulator fixes the problem.

This leak was introduced in version 1.5.0

### DIFF
--- a/src/Turtle/Shell.hs
+++ b/src/Turtle/Shell.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns              #-}
 {-# LANGUAGE CPP                       #-}
 {-# LANGUAGE ExistentialQuantification #-}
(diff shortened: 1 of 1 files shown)


Gabriel439:
I tested this by hand using the following program:

```haskell
import qualified Control.Foldl as Foldl
import qualified Turtle

main :: IO ()
main = do
    x <- Turtle.fold (Turtle.select [(1::Int)..100000000]) Foldl.sum

    print x
```

Before this change it leaked, but after this change it runs in constant space:

Screen Shot 2020-03-28 at 4 36 53 PM

PR #383, opened 2020-05-01
Add pattern synonyms for Size

maoe:
The `Size` type has accessors like `kilobytes :: Integral n => Size -> n` but has no constructors like `KB :: Integral n => n -> Size`. This patch defines such constructors using `PatternSynonyms`. The existing accessors are now defined in terms of record fields of the constructors. 

The constructor names can be controversial. I'm open to suggestions.

The rename of variables (i.e. s/pattern/pattern'/g) is needed because the name `pattern` becomes a keyword when `PatternSynonyms` is enabled.


Gabriel439:
@maoe: I think this is a great idea!  Thank you for contributing this 🙂

PR #384, opened 2020-05-02
Remove an irrelevant doctest comment

maoe:
This doctest section is for the `TB` pattern synonym but this particular example doesn't use `TB` at all. It looks like I forgot to delete these lines when copy-pasting doctest comments in #383.


Gabriel439:
No worries.  Thanks for fixing that 🙂

Issue #388, opened 2020-05-03
`cd` and `export` are global and not per thread


bfrk:
These commands change the global state of the process, so a `cd` or `export` in one thread will have an effect on all other threads as well. This is obvious in retrospect, as these commands can run directly in `IO`, whereas similar libraries like shelly have their own monad transformer to keep this state local. Still, the documentation could make this behavior more explicit as it is quite a pitfall to run into.

Gabriel439:
The fix is up here: https://github.com/Gabriel439/Haskell-Turtle-Library/pull/389

PR #389, opened 2020-05-03
Document that `cd` / `export` have a global affect

Gabriel439:
Fixes https://github.com/Gabriel439/Haskell-Turtle-Library/issues/388

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,12 @@ greatly assist scripting:
 
 * *Modern:* Supports `text` and `system-filepath`
 
+## Caveats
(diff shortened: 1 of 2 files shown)


bfrk:
Thanks! Perhaps you could also add a small note somewhere in the README? This would serve to warn people who e.g. want to port code from a different library where these are local per thread. Which is what happened to me, BTW. Fortunately, I could change our code without too much effort so that it no longer uses `cd` and `export`, but for other applications this might be more difficult.

PR #390, opened 2020-05-06
Add documentation on find command for Unix newcomers

theNerd247:


src/Turtle/Prelude.hs
```diff
@@ -1725,6 +1725,15 @@ inplaceWith sed_ pattern' file = liftIO (runManaged (do
     mv tmpfile file ))
 
 -- | Search a directory recursively for all files matching the given `Pattern`
+-- Note: The `Pattern` matches against the full path of a file as opposed to
+-- just the base name as in GNU find. For example:
+--
+-- > find "foo.txt" "./bar" 
+--
+-- will return only a single filepath: @./bar/foo.txt@. To search for a
```
Gabriel439, line 1733: Are you sure this is correct?  Wouldn't you need to search for the exact string `"./bar/foo.txt"` to match that path?

src/Turtle/Prelude.hs
```diff
@@ -1725,6 +1725,15 @@ inplaceWith sed_ pattern' file = liftIO (runManaged (do
...
+-- filename in a similar manner to GNU find do something similar to:
+--
+-- > find (suffix $ "/" *> "foo.txt") "./bar"
```
Gabriel439, line 1736: What is the `"/"` for?

src/Turtle/Prelude.hs
theNerd247: Oh your right! It was originally `find "foo.txt" "."` and then I went back and updated.

src/Turtle/Prelude.hs
theNerd247: The `"/"` is the file separator for a file path. The intent is to match against `./bar/foo.txt` so `suffix` would match the `./bar` then `"/"` then the file name. I realize though that if one were to have "." as the search-tree root then immediate children ("./foo.txt") wouldn't match. It might have been better to have something like `basename "foo.txt" :: Pattern FilePath"`

Gabriel439: The only reason I asked is because `find` still returns the full name of the path, so I think the closer match to `find` would be something like:

```haskell
find (chars <> "/" <> "foo.txt")
```

PR #393, opened 2020-08-11
Depend on semigroups & fail only on GHC < 8.0

felixonmars:
They are not needed on newer GHC.


Gabriel439:
@felixonmars: Thank you! :slightly_smiling_face:

PR #416, opened 2022-02-06
Add Turtle.Format.makePrinter function

unclechu:
I was trying to implement some abstraction on top of Turtle to redirect the default output (stdout and/or stderr) to a logger function withing some monad where the logger is available. While I was able to implement most of the Turtle function in one way or another it turned out it’s not that trivial to implement custom `printf` and `eprintf` since `(>>-)` of `Turtle.Format.Format` is not exposed.

This change adds `makePrinter` function that just adds extra argument for `printf` that is a printer function ending with `m ()` as `Format` result. The implementation is trivial (`flip (>>-)`), `printf` and `eprintf` are reimplemented using this function.

### DIFF
--- a/src/Turtle/Format.hs
+++ b/src/Turtle/Format.hs
@@ -48,6 +48,7 @@ module Turtle.Format (
     , printf
     , eprintf
     , makeFormat
+    , makePrinter
 
     -- * Parameters
     , w
@@ -107,13 +108,23 @@ instance (a ~ b) => IsString (Format a b) where
 format :: Format Text r -> r
 format fmt = fmt >>- id
 
+{-| Create your own formatted printer function
+
+Can be useful for making custom "printf" and/or "eprintf" functions that would
+use some logger function instead of stdout/stderr within some logger monad.
+
+>>> makePrinter (liftIO . Data.Text.putStr)
+-}
+makePrinter :: (Text -> m ()) -> Format (m ()) r -> r
+makePrinter = flip (>>-)
+
 {-| Print a `Format` string to standard output (without a trailing newline)
 
 >>> printf ("Hello, "%s%"!\n") "world"
 Hello, world!
 -}
 printf :: MonadIO io => Format (io ()) r -> r
-printf fmt = fmt >>- (liftIO . Text.putStr)
+printf = makePrinter (liftIO . Text.putStr)
 
 {-| Print a `Format` string to standard err (without a trailing newline)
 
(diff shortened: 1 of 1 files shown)


src/Turtle/Format.hs
```diff
@@ -107,21 +108,31 @@ instance (a ~ b) => IsString (Format a b) where
 format :: Format Text r -> r
 format fmt = fmt >>- id
 
+{-| Create your own formatted printer function
+
+Can be useful for making custom "printf" and/or "eprintf" functions that would
```
Gabriel439, line 113: This will ensure that the haddocks link to their definitions:

```suggestion
Can be useful for making custom 'printf' and/or 'eprintf' functions that would
```

src/Turtle/Format.hs
```diff
@@ -107,21 +108,31 @@ instance (a ~ b) => IsString (Format a b) where
...
+use some logger function instead of stdout/stderr within some logger monad.
+
+>>> makePrinter (liftIO . Data.Text.putStr)
```
Gabriel439, line 116:
```suggestion
> printf = makePrinter (liftIO . Data.Text.putStr)
```

Gabriel439 (approved):
Looks great!  Just some cosmetic suggestions

src/Turtle/Format.hs
unclechu: @Gabriel439 Referring to the Haddock documentation:

https://haskell-haddock.readthedocs.io/en/latest/markup.html#linking-to-modules

> ### Linking to Modules
>  
> Linking to a module is done by surrounding the module name with double quotes:
>  
> `-- | This is a reference to the "Foo" module.`

unclechu: @Gabriel439 ah, sorry, it’s linking to *modules*, not function names.

unclechu:
@Gabriel439 I have some doubts about the name of the function. Maybe it’s better to explicitly say that it’s about `Format`? Like name the function not just `makePrinter` but `makeFormatPrinter`. What do you think?

Gabriel439:
@unclechu: Another option would be to just expose the `Format` constructor so that people can use `(>>-)` directly.  There's no real reason that it needs to be hidden

unclechu:
@Gabriel439 Yeah, I would like to just get it exposed in the first place so I wouldn’t have this kind of issue.

Gabriel439:
I put  up a PR to expose the `Format` constructor here: https://github.com/Gabriel439/turtle/pull/417

unclechu:
@Gabriel439 I did the same meanwhile https://github.com/Gabriel439/turtle/pull/418 :sweat_smile: Close either one. Anyway, closing this one.

PR #423, opened 2022-04-29
Add latest Win32 support.

bingis-khan:
(and also fix compilation errors when installing turtle with stack.)

- Changed Win32 from <2.9 to >=2.12.
- *Just* changed the call to setFileTime.
- Set LTS version to 19.5 and changed Tested-With GHC 9.0.1 to newer 9.0.2.


Gabriel439:
Thank you! 🙂

PR #439, opened 2022-10-09
mark up license in README.md

das-g:
so that GitHub renders it correctly

(An alternative would be to wrap it in ` ``` ` lines, so that it'd be rendered as a preformatted verbatim ("code") block. But monospace fonts don't make prose very readable, so I'd prefer the change I propose here.)


Gabriella439:
Thank you!

PR #446, opened 2023-06-21
Build against `optparse-applicative-0.18.1.0`

ysangkok:
Tested using `cabal test`. This should allow turtle to get back into Stackage.

PR #447, opened 2023-10-05
Use flag to keep ansi-wl-pprint and optparse-applicative in sync

ysangkok:
This should solve the issues people have been having in #446.


amesgen (approved):
Thanks :heart:

Gabriella439 (approved):
ooooh, this is clever

PR #449, opened 2023-11-01
Update dependencies

erikd:
This is required for to to build it with `ghc-9.8` as a component of a rather large project. 

There are a number of compiler warnings.

I would appreciate a fixed version on Hackage and since this is just a dependency update that could be achieved by a metadata edit.


erikd:
@Gabriella439 Ping?

Gabriella439:
Whoops!  Thanks for the reminder.  I pushed matching revision to Hackage

PR #454, opened 2024-03-02
Support querying environment variable presence/absence

### LINKED ISSUE #452: Check if an environment variable is set
In Bash, it's very common to check if an environment is set and non-empty with `[[ -n "${VAR:-}" ]]`. We should provide helpers for this, like so:

```haskell
hasEnv :: Text -> IO Bool
hasEnv = not . missingEnv

missingEnv :: Text -> IO Bool
missingEnv name = null $ fromMaybe "" $ need name
```

Gabriella439:
Fixes #452

### DIFF
--- a/src/Turtle/Prelude.hs
+++ b/src/Turtle/Prelude.hs
@@ -120,6 +120,10 @@ module Turtle.Prelude (
 #endif
     , need
     , env
+    , isSet
+    , isUnset
+    , isPresent
+    , isMissing
     , cd
     , pwd
     , home
@@ -316,6 +320,7 @@ import Data.IORef (newIORef, readIORef, writeIORef)
 import qualified Data.List as List
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Maybe as Maybe
 import Data.Monoid ((<>))
 import Data.Ord (comparing)
 import qualified Data.Set as Set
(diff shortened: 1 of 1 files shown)


9999years (approved):
Looks good! Two comments:

1. The names don't include `env` or anything like it, which is a little ambiguous, especially because they take a `Text`. I could imagine reading `isPresent "foo"` and thinking it's testing if a file exists.
2. I think `isSet` and `isUnset` are potential footguns — in my experience, it's very rare to consider a variable 'set' if it exists but is set to the empty string. We might consider adding a note to this effect to the documentation. ("In most cases, you want `isPresent`.)

Gabriella439:
I'm happy to add an "env" or something to the names to address (1) so I'll do that.

I'm a bit surprised about (2) because I had the opposite point of view (that `isPresent` / `isAbsent` were the footguns).  The reason why is that treating an in-band value (the empty value) as a sentinel value that means unset is a common source of programming errors (not just in Bash, but in general).  Like, in the case of a simple binary toggle this is less of a concern, but when the environment value stores an actual payload that the payload that the program will use then there is a greater risk that the payload can be meaningfully present, but empty.

9999years:
Hmm, that's true. In my experience most environment variables _are_ used for simple toggles. I would also expect the program to do some validation if the environment variable has semantics (e.g. check that a path exists if it's supposed to represent a path, etc.).

